PR #7, opened 2017-10-06
Expose API function to prepare hash for EIP712 typed data

ukstv:
Exposed function could be used for data validation.


danfinlay:
Looks good, thanks!

PR #10, opened 2017-11-19
initial TweetNaCl-based encryption and decryption

gleim:


gleim:
pending introduction of non-cropped content encryption key material

PR #20, opened 2018-07-13
Implement EIP-712

bitpshr:
This PR introduces a new `signTypedData` method that conforms to the latest version of [EIP-712](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md). The previous `signTypedData` implementation was renamed to `signTypedDataLegacy` and wasn't removed for backwards compatibility.

**Note:** The included test cases and the known values were lifted from the EIP itself for maximum verification.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,6 @@
 const ethUtil = require('ethereumjs-util')
 const ethAbi = require('ethereumjs-abi')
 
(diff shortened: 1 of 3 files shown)


danfinlay:
Published as `eth-sig-util@2.0.0`

PR #22, opened 2018-07-15
EIP-712: Array support

bitpshr:
The original implementation of `signTypedData` omitted array support. This pull request remedies this by adding array support as per [EIP-712](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md) along with updated tests.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -16,7 +16,7 @@ const typedData = {
       ],
       Mail: [
           { name: 'from', type: 'Person' },
-          { name: 'to', type: 'Person' },
+          { name: 'to', type: 'Person[]' },
           { name: 'contents', type: 'string' }
       ],
   },
@@ -32,10 +32,10 @@ const typedData = {
           name: 'Cow',
           wallet: '0xCD2a3d9F938E13CD947Ec05AbC7FE734Df8DD826',
       },
-      to: {
+      to: [{
           name: 'Bob',
           wallet: '0xbBbBBBBbbBBBbbbBbbBbbbbBBbBbbbbBbBbbBBbB',
-      },
+      }],
       contents: 'Hello, Bob!',
   },
 }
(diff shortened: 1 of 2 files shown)


test/index.js
```diff
@@ -318,17 +318,17 @@ test('signedTypeData', (t) => {
   const sig = sigUtil.signTypedData(privateKey, { data: typedData })
 
   t.equal(utils.encodeType('Mail', typedData.types),
-    'Mail(Person from,Person to,string contents)Person(string name,address wallet)')
+    'Mail(Person from,Person[] to,string contents)Person(string name,address wallet)')
```
rmeissner: I would recommend adding an independent test for this, to avoid changing the test values provided in the EIP, this way it is also ensured that the previous behaviour did not break.

bitpshr: Hi @rmeissner, thanks for the feedback. I submitted an upstream PR to update the EIP test values (see ethereum/EIPs#1242.) While we don't plan to land this PR until that upstream PR also lands, you're right that separate tests would be more robust. Will update.

test/index.js
```diff
@@ -16,7 +16,7 @@ const typedData = {
       ],
       Mail: [
           { name: 'from', type: 'Person' },
-          { name: 'to', type: 'Person' },
+          { name: 'to', type: 'Person[]' },
```
rmeissner: when creating the typeHash a lookup is performed based on the type. In this case the type would be `Person[]`, so it would not match `Person` in the types dictionary.

In the test this works since `from` references `Person`, but would be interesting if it still works when we remove that reference.

bitpshr:
Closing this for now and will re-open when EIP-712 array support firms up. [Per the proposal's author](https://github.com/MetaMask/metamask-extension/pull/4803#issuecomment-413391389):

> In the screencast I see that arrays are implemented. This is still somewhat subject to change in EIP712. What I think will happen is that EIP712 will follow whatever encoding Solidity uses internally in memory and/or what ABIv2 uses. This may or may not be the same as what is currently specified. If there is a change, I expect it will be in the handling of fixed size arrays.

PR #31, opened 2018-09-13
Add solidity types to schema

pedrouid:


RonSherfey (changes requested):
requested changes

src/index.ts
```diff
@@ -73,6 +73,24 @@ const TYPED_MESSAGE_SCHEMA = {
   required: ['types', 'primaryType', 'domain', 'message'],
 };
 
+function getSolidityTypes() {
+  const types = ['bool', 'address', 'int', 'uint', 'string', 'byte'];
```
Gudahtt, line 77: This doesn't seem right :thinking: 

From [EIP-712](https://eips.ethereum.org/EIPS/eip-712):
>Definition: The _atomic types_ are `bytes1` to `bytes32`, `uint8` to `uint256`, `int8` to `int256`, `bool` and `address`. These correspond to their definition in Solidity. Note that there are no aliases `uint` and `int`. 
>Definition: The _dynamic types_ are `bytes` and `string`.

Based on that, my interpretation is that `int` and `uint` should not be included, and `byte` should be `bytes`.

```suggestion
  const types = ['bool', 'address', 'string', 'bytes'];
```

I'm not super familiar with the history of this EIP though (I know it went through a few revisions, but I don't know the details), so let me know if this was changed in a more recent revision. Or if I'm just missing something.

Gudahtt (changes requested):
Thanks for the contribution! This looks quite useful. This PR had been around a while unreviewed unfortunately, so I took the liberty of rebasing it onto the main branch and fixing lint errors.

I had have one piece of feedback - let me know what you think!

Gudahtt:
Thanks again for the contribution! I've made a few changes and put this up as a separate PR: #189.

PR #32, opened 2018-09-14
Moving test to jest from tape

godfreyhobbs:
Jest is a bit more standard and gives very good reports.


Gudahtt:
Thanks for the contribution! Unfortunately this PR has gone stale, and it would be more effort than it's worth to rebase it due to the number of changes. We do like the idea of migrating to Jest though. I'll be doing this in a separate PR soon. I'm very sorry about our delayed response - we'll endeavor to reply more quickly from now on.

PR #35, opened 2018-09-14
Adding safely encrypt and decrypt with padding

godfreyhobbs:
Direct use of the `encrypt` method will expose the length of the input message.  The plaintext length can easly be predicted within a few bytes simply by using the ciphertext byte length.

The `encryptSafely` will pad the input into chunks of around 2048 bytes.  Thus the ciphertext length will only expose the nearest 2048 bytse.

Basic test have been added.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,3 +1,4 @@
+const { Buffer } = require('buffer');
 const ethUtil = require('ethereumjs-util');
 const ethAbi = require('ethereumjs-abi');
 const nacl = require('tweetnacl');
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -267,6 +268,42 @@ module.exports = {
...
+    const DEFAULT_PADDING_LENGTH = (2 ** 11);
+    const NACL_EXTRA_BYTES = 16;
+
+    let data = msgParams.data;
+    if (!data) {
+      throw new Error('Cannot encrypt empty msg.data');
+    }
+
+    if (typeof data === 'object' && data.toJSON) {
+      // remove toJSON attack vector
```
danfinlay: I'd love a link to a description of this vector.

test/index.js
```diff
@@ -350,6 +350,39 @@ test("Alice encrypts message with bob's encryptionPublicKey", async t => {
...
+test("Alice encryptsSafely message with bob's encryptionPublicKey", async t => {
+  t.plan(5);
+  const VERSION = 'x25519-xsalsa20-poly1305';
+  const result = await sigUtil.encryptSafely(
+     bob.encryptionPublicKey,
+     secretMessage,
+     VERSION
+  );
+
+  console.log("RESULT", result)
```
danfinlay: Remove extraneous log.

danfinlay:
Would also be good to add docs for this method to the `README`.

Published as `eth-sig-util@2.1.0`.

PR #36, opened 2018-09-24
Add missing const declaration for sanitizedData

### LINKED ISSUE #26: Unnecessary global `sanitizedData`
A unnecessary global variable is introduced in https://github.com/MetaMask/eth-sig-util/blob/master/index.js#L151 (This fails in some mocha test in other projects)

whymarrh:
Fixes #26

This PR adds a missing `const` declaration for the local (previously global) `sanitizeData` used in `sign`. This variable isn't used anywhere else so I think this is safely a small typo.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,21 @@
+version: 2
+
+workflows:
+  version: 2
+  full_test:
+    jobs:
+      - npm-test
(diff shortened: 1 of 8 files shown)


danfinlay:
Seems to have already been merged in. No changes detected.

PR #49, opened 2019-03-19
signTypedDatav3 - ensure strings encoded as utf8

kumavis:
`0x`-prefixed strings were being accidentally interpreted as hex strings. by converting to buffer, we avoid the type ambiguity.


kumavis:
no CI? we should set that up

PR #50, opened 2019-03-19
CI - Add CircleCI v2 config.yml

kumavis:


.circleci/config.yml
```diff
@@ -0,0 +1,21 @@
...
+workflows:
+  version: 2
+  full_test:
+    jobs:
+      - npm-test
+
+jobs:
+  npm-test:
+    docker:
+      - image: circleci/node:8.11.3-browsers
```
bitpshr: Should we match what the extension uses at https://github.com/MetaMask/metamask-extension/blob/develop/.circleci/config.yml#L152?

PR #66, opened 2019-11-10
Cleanup

aakilfernandes:
Looks like this repo has evolved with the various typed signature versions. As someone coming to it with fresh eyes, i thought it could use a cleanup.

#### Adds "Typed Signature Version" explanation to the readme
The copy here is from the metamask docs

#### Combines `signTypedData`, `signTypedData_v3`, and `signTypedData_v4` into a single `signTypedData` method
Rather than various functions, a `typedSignatureVersion` string is passed

#### Combines `recoverTypedSignature `, `recoverTypedSignatureLegacy`, and `recoverTypedSignature a_v4` into a single `recoverTypedSignature` method
Rather than various functions, a `typedSignatureVersion` string is passed

#### Drop of `useV4` flag
The codebase alternates between using a `version` string and a `useV4` flag. `typedSignatureVersion` is used for both

#### Drop default `V4`
Rather than allowing the `typedSignatureVersion` to default to `V4`, it is explicitly required. This should make future versions easier and less error prone.

#### Drop `signTypedMessage`
`signTypedMessage` is redundant with `signTypedData` and is dropped

#### Drop `recoverTypedMessage `
`recoverTypedMessage` is redundant with `recoverTypedData` and is dropped

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,17 @@ Currently there is only one supported signing protocol. More will be added as st
 npm install eth-sig-util --save
 ```
 
+## Typed Signature Versions
+
+| Version | Explanation                                                  |
+| ------- | ------------------------------------------------------------ |
+| `V1`    | This early version of the spec lacked some later security improvements, and should generally be neglected in favor of `V3`. |
+| `V3`    | Currently represents the latest version of the [EIP 712 spec](https://eips.ethereum.org/EIPS/eip-712), making it the most secure method for signing cheap-to-verify data on-chain that we have yet. |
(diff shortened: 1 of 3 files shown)


Gudahtt:
Thanks for the suggestions! These all seem like great ideas to me.

Unfortunately the codebase has changed a lot since this PR was created. In particular the move to TypeScript has made rebasing this difficult. I did start attempting a rebase, but I think it would be easier to re-implement this. I'm going to implement each of these changes as a separate PR.

Gudahtt:
This has been superseded by #156

Issue #59, opened 2019-09-05
Typings don't compile


mrwillis:
There are two issues

```
node_modules/eth-sig-util/index.d.ts:16:5 - error TS2411: Property 'EIP712Domain' of type 'ITypedField[] | undefined' is not assignable to string index type 'ITypedField[]'.

16     EIP712Domain?: Array<ITypedField>
       ~~~~~~~~~~~~
```

I don't recall `EIP712Domain` being optional in the Types in the spec.

Also 

```typescript
export interface ITypedValue {
  [key: string]: string | number | ITypedValue | ITypedValue[]
}
```

should also allow for `string[]` and `number[]`

alcuadrado:
I'm experiencing the same issue. I believe it's impossible to distribute a TS package that has this one as a dependency.

mrwillis:
Yes it is impossible.

danfinlay:
Sorry this has not been a high priority for me at the moment, I'll get to it when I can, or someone else can submit a PR for review.

chentschel:
@danfinlay Its not possible to use the npm with typescript. Can u please review the PR?

danfinlay:
I merged the open PR, fixed one of the two reported issues. Published as `eth-sig-util@2.5.1`

chentschel:
@danfinlay thanks for the merge. As for the 2nd issue @mrwillis , wouldn't it be solved using a ITypedValue[] , defining ITypedValue[] as array of string, or number ? 

```
export interface ITypedValue {
  [key: string]: string | number | ITypedValue | ITypedValue[]
}
```

lazaridiscom:
I'm running too into ts related problems with eth-sig-util, too.

@danfinlay , possibly the index.d.ts file should be dropped for now (see some notes re "I" prefix https://github.com/MetaMask/KeyringController/pull/36#issuecomment-557933069), and the migration to TS should start in lower level packages (see https://github.com/MetaMask/eth-simple-keyring/issues/36). 

@chentschel, if "should also allow for string[] and number[]" is true, then personally I would expect to see those in the declaration file.

danfinlay:
I'm willing to merge a PR either removing `index.d.ts` or improving it.

lazaridiscom:
see #68

lazaridiscom:
(related: MetaMask/metamask-extension#7601)

It looks like `eth-sig-util` should go on top of the migration list.

lazaridiscom:
[...]

PR #67, opened 2019-11-16
Make EIP712Domain non optional. Fix: #59

chentschel:


danfinlay:
Published as `eth-sig-util@2.5.1`

PR #68, opened 2019-11-28
delete the preliminary typedef file

lazaridiscom:
* bump version
* updated package-lock (which was strangely at 2.4.4

(was not sure if i should increase version or not)


lazaridiscom:
(typescript consumers can still use the 2.5.1, until final, ideally autogenerated typedef file is provided)

lazaridiscom:
applied with correct version-no within #69

PR #69, opened 2019-11-30
Version 2.5.2

danfinlay:
Removes erroneous typescript declaration file, pending a proper typescript conversion, per #59.


01SMD (commented):
.

Issue #79, opened 2020-01-18
v4 Release


whymarrh:
This issue tracks the progress of the v3 release

Tasks:

- [x] Convert to TypeScript (#70, #74)
- [x] Add linting (#83)
- [x] Confirm Node 12 works/fix it if it doesn't (#72)
- [x] Update README with publishing instructions
- [ ] Update package to be under `@metamask` scope
- [x] Switch from `npm` to `yarn` (#82)

KiChjang:
I got a question: as I was working on #74, I was thinking "man, it'd be so great if the type of `MessageTypes` could be changed from
```typescript
interface MessageTypeProperty {
  name: string;
  type: string;
}

interface MessageTypes {
  EIP712Domain: MessageTypeProperty[];
  [additionalProperties: string]: MessageTypeProperty[];
}
```
to
```typescript
interface MessageTypeProperty {
  [name: string]: string;
}

interface MessageTypes {
  EIP712Domain: MessageTypeProperty;
  [additionalProperties: string]: MessageTypeProperty;
}
```

The consequences of this is profound, because it would let me type `TypedMessage` like so:

```typescript
export interface TypedMessage<T extends MessageTypes, K extends keyof T> {
  types: T;
  primaryType: K;
  domain: {
    name?: string
    version?: string
    chainId?: number
    verifyingContract?: string
  };
  message: { [P in T[K]]: any };
}
```

What did I just do there? I ensured that the `message` property matches exactly the type that you passed into `primaryType`, and also ensured that the fields in `message` appears as you specified in `types`. However, this necessitated an API change, and I was uncertain whether it'd be a good change until I saw that this module will be going through a major version bump.

Would this be a desirable change? In summary, here's a one-liner description of what needs to change:
* Accept an object containing the names of the fields as fields in `MessageTypeProperty`, instead of it being an array of objects containing a `name` field and a `type` field

KiChjang:
In fact, I can continue to type this further by typing the EIP 712 types:
```typescript
type EIP712DataType = AtomicType | DynamicType | ReferenceType;
```

`ReferenceType` would be a little bit tricky since it will have to reference the name of the types as defined in the `types` field of `TypedMessage`, but it would be great to experiment and see how far we can push the type system to ensure absolute correctness and conformity to EIP 712.

and then ensuring that `MessageTypeProperty` contains actual EIP 712-defined data types:

```typescript
interface MessageTypeProperty {
    [name: string]: EIP712DataType;
[...]

PR #83, opened 2020-01-23
Add ESLint config

whymarrh:
Refs #79

This PR adds an ESLint config.


whymarrh:
I'm leaving this in draft while I tinker with the shared config.

whymarrh:
@Gudahtt @rekmarks this should be good to review now

01SMD (commented):
.

Issue #85, opened 2020-03-23
Signing Verification appears to behave differently in browser than in NodeJS v12


pmumby:
Currently I have a use-case for an off-chain application to verify ownership of an address (specifically a Discord Bot).

In this use-case, the bot gathers some parameters, sends the user to the URL of a webpage running a pure JS dapp which uses Metamask to sign a message. It then gives them a string to paste back into the bot, which the bot uses via the same JS code on the back-end to verify the signature.

When I do this, in the browser my code generates the message from 4 parameters, id, name, datestamp, and address.

When I pass those parameters to this code in browser, and sign, metamask pops up, everything looks good, signature works, and the verification code is able to verify the signature is valid just fine.

When I take the same signature, and same parameters and pass it to the exact same code running on NodeJS v12, the verification fails. The address it recovers is completely different.

I've attempted this with the Personal.sign, signTyped, and signTypedV3, all behave the same.

It is expected the behavior of the off-chain verification code would be the same regardless of which JS runtime it's executing in.

The following is code which reproduces the issue for me:
```
const Signer = {
    EIP712DomainSchema: [
        { name: "name", type: "string" },
        { name: "version", type: "string" },
        { name: "salt", type: "bytes32" },
    ],
    DiscordAuthorizationSchema: [
        { name: "statement", type: "string" },
        { name: "id", type: "uint256" },
        { name: "name", type: "string" },
        { name: "date", type: "string" },
        { name: "address", type: "address" },
    ],
    domainData: {
        name: "Discord Address Authorization",
        version: "1.0",
        salt: "0xe2d857f4a3edcb9b78b4d503120433db1e3f6cdc2b7971ab739626c97e86a558"
    },
    buildStatementString: function (id, name, date, address) {
        return (
            "I Hereby verify that I am the owner of the Ethereum Address: " + address + "\n" +
            "I authorize the use of this address by the Discord User ID: " + id + " Named: " + name + "\n" +
            "Signed on Date: " + date
        )
    },
    buildMessageData: function (id, name, date, address) {
        return {
            statement: this.buildStatementString(id, name, date, address),
            id: id,
            name: name,
            date: date,
            address: address
        }
    },
    buildSigningPacket: function (id, name, date, address) {
        return JSON.stringify({
            types: {
                EIP712Domain: this.EIP712DomainSchema,
                DiscordAuthorization: this.DiscordAuthorizationSchema
            },
            domain: this.domainData,
[...]

PR #89, opened 2020-04-27
ci - test against multiple node versions

kumavis:
https://github.com/MetaMask/eth-sig-util/issues/85 suggests we're signing different on different node versions.

This PR runs CI against different node versions and chrome

this PR does not reproduce the issue in #85 
a test for that should be added when the details are known

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,10 @@ workflows:
             - prep-deps
       - test-14:
           requires:
-            - prep-deps                  
+            - prep-deps
+      - test-browser:
+          requires:
+            - prep-deps
 
 
 jobs:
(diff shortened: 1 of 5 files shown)


kumavis:
need to test in browser or at least a browserify bundle

kumavis:
added browser tests via `browserify` and `smokestack`
took too long grumble grumble

kumavis:
do we need to fan in to a all-tests-pass for an easier github required test check?

whymarrh:
> do we need to fan in to a all-tests-pass for an easier github required test check?

Yes please, I was hoping to add that to the workflow at some point but it fell off my radar before we started using this same workflow in a few places

kumavis:
@whymarrh did the fan in and set `all-tests-pass` as a github expected CI check

PR #104, opened 2020-09-24
rename TypedDataUtils.sign to TypedDataUtils.eip712Hash

ashwinYardi:


Gudahtt (approved):
Thanks for the contribution! I think this is a great idea.

This is a breaking change, but we'll be preparing a new major release soon, so this is OK.

src/index.ts
```diff
@@ -292,10 +292,10 @@ const TypedDataUtils = {
   /**
    * Signs a typed message as per EIP-712 and returns its keccak hash
```
Gudahtt, line 293: Some of these other inline docs could use improvements too, but I can do that in a separate PR

PR #111, opened 2021-01-08
Standardize TypeScript and lint configuration

rekmarks:
- TypeScript configuration
  - Update target from `es5` to `ES2017`
  - Add source maps and other config options per other repositories
  - Add multiple `tsconfig` files to support separate bundling of source and test files
- Move to `src`/`dist` file structure
- Update `eslint` config and dependencies
  - Manually re-type a couple of things to fix lint errors
- Run `prettier` on everything

No functional changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,3 @@ package-lock.json
 
 dist
 test/*.js
-src/*.js
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,10 @@
-node_modules
 .DS_Store
+node_modules
 package-lock.json
 
 *.tgz
+*.log
 
-# TypeScript
-/build
-**/*.d.ts
-**/*.js
+dist
+test/*.js
+src/*.js
--- a/package.json
+++ b/package.json
@@ -35,13 +35,13 @@
     "tweetnacl-util": "^0.15.0"
   },
   "devDependencies": {
-    "@metamask/eslint-config": "^4.1.0",
-    "@types/node": "^10.17.13",
-    "@typescript-eslint/eslint-plugin": "^4.12.0",
-    "@typescript-eslint/parser": "^4.12.0",
+    "@metamask/eslint-config": "^5.0.0",
+    "@types/node": "^14.14.25",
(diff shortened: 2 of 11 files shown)


Gudahtt (changes requested):
I've removed `eccrypto-lite` in a separate PR: https://github.com/MetaMask/eth-sig-util/pull/116

Once that is merged, could you rebase this to fix the conflicts?

rekmarks:
@Gudahtt done!

package.json
```diff
@@ -42,15 +41,17 @@
...
     "@types/node": "^10.17.13",
-    "@typescript-eslint/eslint-plugin": "^2.20.0",
-    "@typescript-eslint/parser": "^2.20.0",
+    "@typescript-eslint/eslint-plugin": "^4.12.0",
+    "@typescript-eslint/parser": "^4.12.0",
     "browserify": "^16.5.1",
     "eslint": "^6.8.0",
-    "eslint-plugin-import": "^2.20.1",
+    "eslint-plugin-import": "^2.22.1",
+    "eslint-plugin-json": "^2.1.2",
```
Gudahtt, line 51: We should consider dropping this in favour of `prettier`. That's what we did on the extension.

.gitignore
```diff
@@ -1,12 +1,9 @@
-test/bundle.js
```
Gudahtt, line 1: This still needs to be ignored until #115 is merged. We could merge that first though.

package.json
rekmarks: I'd like to see us add `prettier` to our base `eslint` config first, and add the `prettier` plugin as a peerDependency. In fact, I was discussing this with @brad-decker earlier, and I also suggested that we do this: https://github.com/MetaMask/eslint-config/issues/104

Gudahtt: Oh I meant specifically for linting `.json` files. It was added to the extension to lint JSON before it was used for JavaScript.

But sure, we can wait. That works too.

Gudahtt (approved):
LGTM!

PR #115, opened 2021-02-04
Switch from `smokestack` to `airtap`

### LINKED ISSUE #109: Replace browser test tools
Our current browser test tools are niche and unmaintained, and should be replaced with something more modern.

Gudahtt:
Fixes #109

We are now using `airtap` for running unit tests in the browser, rather than `smokestack`. `smokestack` is no longer actively maintained, and has dependencies with various security vulnerabilities. `airtap` is actively maintained and much easier to use.

### DIFF
--- a/.circleci/scripts/deps-install.sh
+++ b/.circleci/scripts/deps-install.sh
@@ -6,6 +6,3 @@ set -u
 set -o pipefail
 
 yarn --frozen-lockfile --ignore-scripts --har
-
-# install script whitelist
-(cd node_modules/wd && yarn run install)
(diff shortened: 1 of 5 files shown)


.circleci/scripts/deps-install.sh
```diff
@@ -8,4 +8,3 @@ set -o pipefail
 yarn --frozen-lockfile --ignore-scripts --har
 
 # install script whitelist
```
darkwing, line 10: I'm guessing we also don't need this comment?

Gudahtt: Removed!

rekmarks (approved):
LGTM!

PR #144, opened 2021-07-09
Update ESLint config to v7.0.1

Gudahtt:
The ESLint config has been updated, which includes the addition of `prettier. All lint issues have been fixed with `yarn lint:fix`. The ignore file was replaced by the ignore patterns property as well.

The Prettier plugin `prettier-plugin-packagejson` has been added to sort `package.json`.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,17 +1,10 @@
 module.exports = {
   root: true,
-  extends: [
-    '@metamask/eslint-config',
-    '@metamask/eslint-config/config/nodejs',
-    '@metamask/eslint-config/config/typescript',
-  ],
-  plugins: ['json'],
+  extends: ['@metamask/eslint-config', '@metamask/eslint-config-nodejs'],
   overrides: [
     {
-      files: ['*.js', '*.json'],
-      parserOptions: {
-        sourceType: 'script',
-      },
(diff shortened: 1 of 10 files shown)


.eslintrc.js
```diff
@@ -27,5 +20,5 @@ module.exports = {
     ],
     'no-param-reassign': 'off',
   },
-  ignorePatterns: ['!.eslintrc.js', './test/*.js', 'dist'],
+  ignorePatterns: ['!.eslintrc.js', 'test/*.js', 'dist'],
```
shanejonas, line 23: should we add back the `coverage` and `*.d.ts` from the removed `.eslintignore` . And the `node_modules` ignore should maybe go in `tsconfig`?

Gudahtt: Ah, good catch! But it looks like all of these are fine to ignore. `node_modules` is [ignored by default](https://eslint.org/docs/user-guide/configuring/ignoring-code#the-eslintignore-file), and we don't have any `coverage` directory or `*.d.ts` files in this repo.

Gudahtt:
Just rebased this to fix lint errors introduced when the branch was updated.

rekmarks (approved):
LGTM!

PR #147, opened 2021-07-13
Add `@lavamoat/allow-scripts`

Gudahtt:
`@lavamoat/allow-scripts` is now used to control which post-install scripts are run. Any required solely by dev dependencies were disabled.

A new `setup` command has been added to make it easier to run allow-scripts after each install. The README has been updated with our standard contributor instructions, which includes an explanation of this `setup` command.

PR #148, opened 2021-07-13
Migrate from CircleCI to GitHub Actions

Gudahtt:
All CI checks have been migrated to GitHub actions, using our MetaMask module template example workflow. The main difference between this and the template workflow is the addition of the `test:browser` command for running browser tests.

This should work similarly to the CircleCI workflow. The main differences are that now unit tests, the linter, and browser tests are run with all three supported Node.js versions (previously they were run with Node.js v12, except unit tests which were also run with v14). Also we now have dependency caching, so hopefully the workflows will be faster. The HAR archive from installation has been omitted as well.


Gudahtt:
This depends upon #147

Gudahtt:
Note that the build failure is just from CircleCI, which has had its config removed in this branch. I can disable CircleCI status updates and update the required status updates in our branch protection rules once this has been approved.

rekmarks (approved):
We could (probably should?) separate building from testing at the script level and note that the GitHub workflow builds as well as lints and tests, but I won't block on it!

Gudahtt:
Oh, good idea! I'll do that in a follow up PR.

Gudahtt:
Actually, on second thought, it would be challenging to split them apart. If we do that, we risk `yarn test` blowing up with a confusing error message when the build doesn't exist. Or even worse, it could run using an outdated build. We could prevent that by checking to see if there was a build first, but I'm not sure it's worth the effort.

I'll soon be converting these tests to Jest, then migrating them to TypeScript. At that point, the special test bundle and the build step will no longer be required for testing, so they will be removed. Maybe we can just wait until then to clean all of this up.

PR #152, opened 2021-07-14
Allow `TypedDataUtils` to be called unbound

Gudahtt:
Previously the functions exposed as `TypedDataUtils` could only be called from the `TypedDataUtils` object. If they were called unbound, they would throw errors because of the use of `this`.

They have all been updated to no longer rely upon `this`, so they now work the same way regardless how they are bound when called.

They are still exported as the `TypedDataUtils` object, so this should not change the API. This was done to simplify the code, specifically to make it easier for functions outside of `TypedDataUtils` to reuse code inside of `TypedDataUtils`.

Some types required adjustments, as type mistakes were brought to light that TypeScript for some reason wasn't aware of when these were declared as properties of the `TypedDataUtils` object. These were fixed by adding two type assertions, making the `types` parameter to `hashStruct` and `hashType` more strict, and by making the `typedData` parameter to `eip712Hash` more strict. The type assertions (warranted or not) preserve the types used previously. We can replace them later with validation.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -74,256 +74,273 @@ const TYPED_MESSAGE_SCHEMA = {
 };
 
 /**
- * A collection of utility functions used for signing typed data
+ * Encodes an object by encoding and concatenating each of its members
+ *
+ * @param {string} primaryType - Root type
+ * @param {Object} data - Object to encode
+ * @param {Object} types - Type definitions
+ * @param {Version} version - The EIP-712 version the encoding should comply with
+ * @returns {Buffer} - Encoded representation of an object
  */
-const TypedDataUtils = {
-  /**
-   * Encodes an object by encoding and concatenating each of its members
-   *
-   * @param {string} primaryType - Root type
-   * @param {Object} data - Object to encode
-   * @param {Object} types - Type definitions
-   * @param {Version} version - The EIP-712 version the encoding should comply with
-   * @returns {Buffer} - Encoded representation of an object
-   */
-  encodeData(
-    primaryType: string,
-    data: Record<string, unknown>,
-    types: Record<string, MessageTypeProperty[]>,
-    version: Version,
-  ): Buffer {
-    const encodedTypes = ['bytes32'];
-    const encodedValues = [this.hashType(primaryType, types)];
+function encodeData(
+  primaryType: string,
+  data: Record<string, unknown>,
+  types: Record<string, MessageTypeProperty[]>,
+  version: Version,
+): Buffer {
+  const encodedTypes = ['bytes32'];
+  const encodedValues: unknown[] = [hashType(primaryType, types)];
 
-    if (version === 'V4') {
-      const encodeField = (name, type, value) => {
-        if (types[type] !== undefined) {
-          return [
-            'bytes32',
-            value == null // eslint-disable-line no-eq-null
-              ? '0x0000000000000000000000000000000000000000000000000000000000000000'
-              : ethUtil.keccak(this.encodeData(type, value, types, version)),
-          ];
+  if (version === 'V4') {
+    const encodeField = (name, type, value) => {
+      if (types[type] !== undefined) {
+        return [
+          'bytes32',
+          value == null // eslint-disable-line no-eq-null
+            ? '0x0000000000000000000000000000000000000000000000000000000000000000'
+            : ethUtil.keccak(encodeData(type, value, types, version)),
+        ];
+      }
+
+      if (value === undefined) {
+        throw new Error(`missing value for field ${name} of type ${type}`);
+      }
+
+      if (type === 'bytes') {
+        return ['bytes32', ethUtil.keccak(value)];
+      }
+
+      if (type === 'string') {
+        // convert string to buffer - prevents ethUtil from interpreting strings like '0xabcd' as hex
+        if (typeof value === 'string') {
+          value = Buffer.from(value, 'utf8');
         }
+        return ['bytes32', ethUtil.keccak(value)];
+      }
 
-        if (value === undefined) {
-          throw new Error(`missing value for field ${name} of type ${type}`);
-        }
+      if (type.lastIndexOf(']') === type.length - 1) {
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -74,253 +74,272 @@ const TYPED_MESSAGE_SCHEMA = {
...
+ *
+ * @param {string} primaryType - Root type
+ * @param {Object} data - Object to hash
+ * @param {Object} types - Type definitions
+ * @returns {Buffer} - Hash of an object
+ */
+function hashStruct(
+  primaryType: string,
+  data: Record<string, unknown>,
+  types: Record<string, MessageTypeProperty[]>,
```
Gudahtt, line 257: Here is the parameter type adjustment that was required to eliminate type errors. Clearly this is required because it's passed straight to `encodeData`, which itself required the `types` parameter to be of this type.

src/index.ts
```diff
@@ -74,253 +74,272 @@ const TYPED_MESSAGE_SCHEMA = {
...
+          value = ethUtil.keccak(value);
+          encodedValues.push(value);
+        } else if (types[field.type] !== undefined) {
+          encodedTypes.push('bytes32');
+          value = ethUtil.keccak(
+            encodeData(
+              field.type,
+              // TODO: Add validation to ensure this is a string-indexed
+              // object, so that this type cast can be removed
+              value as Record<string, unknown>,
```
Gudahtt, line 171: Here's the first type assertion that was required. This should be guaranteed to be a `Record<string, any>` for any properly formatted message, but we don't have sufficient type information at this point to guarantee that. We can add that later.

src/index.ts
```diff
@@ -74,253 +74,272 @@ const TYPED_MESSAGE_SCHEMA = {
...
+    ),
+  );
+  if (sanitizedData.primaryType !== 'EIP712Domain') {
     parts.push(
-      this.hashStruct(
-        'EIP712Domain',
-        sanitizedData.domain,
+      hashStruct(
+        // TODO: Validate that this is a string, so this type cast can be removed.
+        sanitizedData.primaryType as string,
```
Gudahtt, line 322: Here's the second type assertion that was required. At first it looked like this was already guaranteed to be a string, because `primaryType` is a keyof the `MessageTypes` object, and that's string-indexed. But apparently TypeScript is screwy when it comes to index types, and it happily allows using a `number` index as well (I tested it to be sure), despite this attempt at preventing that. So technically this could be a number. But that's clearly against the spec, and it's clearly what we were trying to prevent with that type.

We can shore this up later with runtime validation. Or better types. Or both.

test/index.ts
```diff
@@ -1389,3 +1389,122 @@ test('signedTypeMessage V4 with recursive types', (t) => {
     '0xf2ec61e636ff7bb3ac8bc2a4cc2c8b8f635dd1b2ec8094c963128b358e79c85c5ca6dd637ed7e80f0436fe8fce39c0e5f2082c9517fe677cc2917dcd6c84ba881c',
   );
 });
+
+test('unbound sign typed data utility functions', (t) => {
```
Gudahtt, line 1393: I tested that these all failed on the `main` branch.

Gudahtt:
I'd strongly recommend hiding whitespace when looking at this diff. It's a mess when whitespace is shown because everything was de-indented.

src/index.ts
```diff
@@ -74,253 +74,272 @@ const TYPED_MESSAGE_SCHEMA = {
...
-          ];
-        }
+function encodeData(
+  primaryType: string,
+  data: Record<string, unknown>,
+  types: Record<string, MessageTypeProperty[]>,
+  useV4 = true,
+): Buffer {
+  const encodedTypes = ['bytes32'];
+  const encodedValues: unknown[] = [hashType(primaryType, types)];
```
Gudahtt, line 91: This was another type change that was required. TypeScript assumed that this was supposed to be a `Buffer` array because the first entry was a `Buffer`. But it looks like it's just an array of unknown data, not just of buffers.

src/index.ts
```diff
@@ -74,253 +74,272 @@ const TYPED_MESSAGE_SCHEMA = {
...
+/**
+ * Hashes the type of an object
+ *
+ * @param {string} primaryType - Root type to hash
+ * @param {Object} types - Type definitions
+ * @returns {Buffer} - Hash of an object
+ */
+function hashType(
+  primaryType: string,
+  types: Record<string, MessageTypeProperty[]>,
```
Gudahtt, line 272: This parameter type changed for the same reason as `hashStruct`

src/index.ts
```diff
@@ -74,253 +74,272 @@ const TYPED_MESSAGE_SCHEMA = {
...
-    const sanitizedData = this.sanitizeData(typedData);
-    const parts = [Buffer.from('1901', 'hex')];
+/**
+ * Signs a typed message as per EIP-712 and returns its keccak hash
+ *
+ * @param {Object} typedData - Types message data to hash as per eip-712
+ * @returns {Buffer} - keccak hash of the resulting signed message
+ */
+function eip712Hash<T extends MessageTypes>(
+  typedData: TypedData | TypedMessage<T>,
```
Gudahtt, line 305: This was never actually passed in `Partial<>` data. Also `sanitizeData` requires that the typed data be complete.

Gudahtt:
Just rebased this onto `main` and updated the commit message. No other changes made.

shanejonas:
Havent touched this repo yet, but the changes to typings look good, +1 for removal of `this`.

shanejonas (dismissed):

Gudahtt:
Thanks for the review @shanejonas! Just had to rebase this to resolve conflicts. The diff should be identical now except for the tests, which now need "V4" specified explicitly in a few places due to the consolidation PR that was merged.

test/index.ts
shanejonas: will need to be converted to `jest` style to go along with this PR: https://github.com/MetaMask/eth-sig-util/pull/161

Gudahtt: Done!

shanejonas (approved):

PR #156, opened 2021-07-15
Consolidate `signTypedData` and `recoverTypedSignature` functions

Gudahtt:
The `signTypedData` and `recoverTypedSignature` functions each had three different implementations (one for each version of EIP-712). They have each been consolidated into a single function each.

The functions `signTypedData` and `recoverTypedSignature` used to be specifically for `eth_signTypedData_v3`. Instead they now accept an additional `version` parameter that specifies which version should be used. This replaces `signTypedDataLegacy` and `recoverTypedSignatureLegacy`, which were for `eth_signTypedData_v1`, and it replaces `signTypedData_v4` and `recoverTypedSignature_v4`.

The functions `signTypedMessage` and `recoverTypedMessage` used to provide a single interface for using any of EIP-712 types. But now that the base functions serve that purpose instead, they are now obsolete and have been removed.

Additionally a few `TypedDataUtils` functions were updated to accept a `version` parameter instead of a `useV4` parameter, to bring them in-line with the other functions. The functions changed are `hashStruct`, `eip712Hash`, and `encodeData`.

The README and inline documentation has been updated accordingly. New doc strings were added for `signTypedData` and `recoverTypedSignature`, which didn't have any yet before now.

Credit to @aakilfernandes for this idea, who first implemented this in PR #66.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,39 +36,21 @@ msgParams should have a `data` key that is hex-encoded data unsigned, and a `sig
 
 Returns a hex-encoded sender address.
 
-### signTypedDataLegacy (privateKeyBuffer, msgParams)
+### signTypedData (privateKeyBuffer, msgParams, version)
 
-Signs typed data as per [an early draft of EIP 712](https://github.com/ethereum/EIPs/pull/712/commits/21abe254fe0452d8583d5b132b1d7be87c0439ca).
-
-Data should be under `data` key of `msgParams`. The method returns prefixed signature.
-
-### signTypedData (privateKeyBuffer, msgParams)
-
-Signs typed data as per [EIP 712](https://github.com/ethereum/EIPs/pull/712), except that arrays and recursive data structures are not supported.
+Sign typed data according to EIP-712. The signing differs based upon the `version`, which is explained in the table below.
 
 Data should be under `data` key of `msgParams`. The method returns prefixed signature.
 
-### signTypedData_v4 (privateKeyBuffer, msgParams)
+| Version | Description                                                  |
+| ------- | ------------------------------------------------------------ |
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -36,39 +36,21 @@ msgParams should have a `data` key that is hex-encoded data unsigned, and a `sig
...
 
-Signs typed data as per [an early draft of EIP 712](https://github.com/ethereum/EIPs/pull/712/commits/21abe254fe0452d8583d5b132b1d7be87c0439ca).
+Sign typed data according to EIP-712. The signing differs based upon the `version`, which is explained in the table below.
 
 Data should be under `data` key of `msgParams`. The method returns prefixed signature.
 
-### signTypedData (privateKeyBuffer, msgParams)
+| Version | Description                                                  |
+| ------- | ------------------------------------------------------------ |
+| `V1`    | This is based on [an early version of EIP-712](https://github.com/ethereum/EIPs/pull/712/commits/21abe254fe0452d8583d5b132b1d7be87c0439ca) that lacked some later security improvements, and should generally be neglected in favor of `V3`. |
```
shanejonas, line 47: should we add a note about the lack of `V2`?

shanejonas (approved):

README.md
Gudahtt: Probably, yep! I just didn't know what to say - I have no idea why there isn't a V2

Gudahtt: Oh, actually, I just found more information on this in our docs:

>The missing `v2` represents an intermediary design that was implemented by the Cipher browser, so that we have room to implement it if there is ever enough developer demand for it.

I'll add this in a follow-up PR.

PR #159, opened 2021-07-15
Use standard TypeScript config

### LINKED ISSUE #142: Include original .ts files in the NPM package to allow proper debugging
The published package in NPM only contains the `/dist/*` files but when debugging with i.e. the Chrome DevTools and stepping into the library, an error is thrown because the original `/src/index.ts` file cannot be found and therefore the sourcemaps can't be used to display the proper source lines being executed.

Please update the NPM package to also include the `/src` directory.

Note: To workaround this issue, just copied the `/src` file into `node_modules/eth-sig-util` and the DevTools are happy again.

Gudahtt:
The TypeScript config has been updated to match the configuration we use in other libraries, mostly. The remaining differences are that `strict` is still not enabled, and there is a separate config used for testing. The test config was preserved for now, but that will be replaced soon when we migrate the tests to Jest. `strict` mode will be enabled in a later PR after we update dependencies, as that will make the necessary changes much easier.

The changes are:
* `inlineSources` has been enabled. This embeds a complete copy of the source code with the source maps, allowing for a much better debugging experience.
* `lib` has been set to `ES2020`, ensuring that we only rely upon JavaScript APIs. Previously it also allowed use of browser APIs.
* `rootDir` is now set explicitly to `src`, which will help ensure we don't accidentally change the structure of the package.
* The old `prod` config was merged with the base config, so it was no longer needed as a separate file.

Closes #142

### DIFF
--- a/tsconfig.test.json
+++ b/tsconfig.test.json
@@ -2,7 +2,10 @@
   "extends": "./tsconfig.json",
   "compilerOptions": {
     "declaration": false,
+    "inlineSources": false,
+    "rootDir": "",
+    "outDir": "",
     "sourceMap": false
   },
-  "files": ["./test/index.ts"]
+  "include": ["./test/index.ts"]
 }
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   ],
   "scripts": {
     "setup": "yarn install && yarn allow-scripts",
-    "build": "tsc --project ./tsconfig.prod.json && tsc --project ./tsconfig.test.json",
+    "build": "tsc --project ./tsconfig.json && tsc --project ./tsconfig.test.json",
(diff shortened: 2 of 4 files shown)


tsconfig.test.json
```diff
@@ -2,7 +2,10 @@
   "extends": "./tsconfig.json",
   "compilerOptions": {
     "declaration": false,
+    "inlineSources": false,
+    "rootDir": "",
+    "outDir": "",
```
rekmarks, line 7: Do we output the test `index.js` to the repo root?

Gudahtt: Nope! Seemingly this "unsets" the `outDir` set in the base config. So it defaults to outputting it alongside the source, in the `test` directory.

rekmarks (approved):
LGTM!

PR #161, opened 2021-07-16
Migrate tests from tape to Jest

Gudahtt:
The unit tests have been migrated from tape to Jest. The standard Jest config from the MetaMask module template has been used.

We're using `ts-jest` to compile the tests and run them all at once, so we don't need a build step for the tests now either. The `build` command is now run separately on CI.

The tape tests didn't break the tests into sections, which is commonly done for our Jest tests. So `describe` blocks haven't yet been used. I will be adding `describe` sections in later PRs though, as the tests are made more comprehensive.

The tests were left mostly functionally equivalent. Some included minor changes, like checking the length of a property rather than just whether it's truthy, or checking the error message, to confirm with our Jest lint rules.


Gudahtt:
Depends upon #156

shanejonas (approved):

PR #164, opened 2021-07-27
Add encode data tests

Gudahtt:
The function `TypedDataUtils.encodeData` has been thoroughly tested with all novel inputs I could think of. Each supported version (V3 and V4) have been tested independently with each input. There are also tests for each input that should have an identical encoding between V3 and V4, and each that should have non-matching signatures.

The structure of the tests is explained in a large inline comment at the start of the `describe` block for `TypedDataUtils.encodeData`.

The `encodeData` assertions that accompanied the signature tests have been deleted, as they're now redundant.

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -1,6 +1,1431 @@
 import * as ethUtil from 'ethereumjs-util';
 import * as sigUtil from '.';
 
+const encodeDataExamples = {
+  // dynamic types supported by EIP-712:
+  bytes: [10, '10', '0x10', Buffer.from('10', 'utf8')],
+  string: [
+    'Hello!',
+    '0xbBbBBBBbbBBBbbbBbbBbbbbBBbBbbbbBbBbbBBbB',
+    '0xabcd',
+    '😁',
+    10,
+  ],
+  // atomic types supported by EIP-712:
+  address: [
+    '0xbBbBBBBbbBBBbbbBbbBbbbbBBbBbbbbBbBbbBBbB',
+    '0x0',
+    10,
+    'bBbBBBBbbBBBbbbBbbBbbbbBBbBbbbbBbBbbBBbB',
+    Number.MAX_SAFE_INTEGER,
+  ],
+  bool: [true, false, 'true', 'false', 0, 1, -1, Number.MAX_SAFE_INTEGER],
+  bytes1: [
+    '0x10',
+    10,
+    0,
+    1,
+    -1,
+    Number.MAX_SAFE_INTEGER,
+    Buffer.from('10', 'utf8'),
+  ],
+  bytes32: [
+    '0x10',
+    10,
+    0,
+    1,
+    -1,
+    Number.MAX_SAFE_INTEGER,
+    Buffer.from('10', 'utf8'),
+  ],
+  int8: [0, '0', '0x0', 255, -255],
+  int256: [0, '0', '0x0', Number.MAX_SAFE_INTEGER, Number.MIN_SAFE_INTEGER],
+  uint8: [0, '0', '0x0', 255],
+  uint256: [0, '0', '0x0', Number.MAX_SAFE_INTEGER],
+  // atomic types not supported by EIP-712:
+  int: [0, '0', '0x0', Number.MAX_SAFE_INTEGER, Number.MIN_SAFE_INTEGER], // interpreted as `int256` by `ethereumjs-abi`
+  uint: [0, '0', '0x0', Number.MAX_SAFE_INTEGER], // interpreted as `uint256` by `ethereumjs-abi`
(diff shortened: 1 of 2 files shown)


Gudahtt:
This depends upon #152 and #161

Gudahtt:
This PR has been updated with more inline comments explaining what the tests are for.

src/index.test.ts
```diff
@@ -1,6 +1,1431 @@
...
+        ],
+      };
+      const primaryType = 'Mail';
+      const message = {
+        from: {
+          name: 'Cow',
+          wallet: '0xCD2a3d9F938E13CD947Ec05AbC7FE734Df8DD826',
+        },
+        contents: 'Hello, Bob!',
+      };
```
rekmarks, line 536: This test case calls from a custom type value with a property "set to undefined", whereas the property in question (`to`) is actually not there at all. The test would be slightly more readable if the message object included `to: undefined`. It would also match what you've done in other test cases.

src/index.test.ts
```diff
@@ -1,6 +1,1431 @@
...
+          message,
+          types,
+          'V3',
+        ).toString('hex'),
+      ).toMatchSnapshot();
+    });
+  });
+
+  // This test suite should include all expected V4 behavior except the example
+  // data encoding, whcih is covered in the "V3/V4 encoding similarities" test
```
rekmarks:
```suggestion
  // data encoding, which is covered in the "V3/V4 encoding similarities" test
```

src/index.test.ts
```diff
@@ -1,6 +1,1431 @@
...
+// Union of all types from both sets of examples
+const allExampleTypes = [
+  ...new Set(
+    Object.keys(encodeDataExamples).concat(
+      Object.keys(encodeDataErrorExamples),
+    ),
+  ),
+];
+
+describe('TypedDataUtils.encodeData', function () {
```
rekmarks, line 82: Why are we not using arrow functions for the callbacks in this test suite? `mocha` hangover?

src/index.test.ts
rekmarks: The same goes for the corresponding test case in the V4 suite.

src/index.test.ts
```diff
@@ -1,6 +1,1431 @@
...
+  // | Unrecognized primary type                            | N  | N  |                     |
+  // | Unrecognized non-primary type                        | N  | N  |                     |
+  // | Extra type specified that isn't used by primary type | Y  | Y  | Y                   |
+
+  describe('V3', function () {
+    describe('example data', function () {
+      // Reassigned to silence "no-loop-func" ESLint rule
+      // It was complaining because it saw that `it` and `expect` as "modified variables from the outer scope"
+      // which can be dangerous to reference in a loop. But they aren't modified in this case, just invoked.
+      const _expect = expect;
```
Gudahtt, line 128: I really wish there was some way to silence this false positive. I didn't want to disable the rule for the whole file because it's genuinely useful still, but this is irritating.

src/index.test.ts
```diff
@@ -1,6 +1,1431 @@
...
+          types,
+          'V3',
+        ).toString('hex'),
+      ).toMatchSnapshot();
+    });
+  });
+
+  // This test suite should include all expected V4 behavior except the example
+  // data encoding, whcih is covered in the "V3/V4 encoding similarities" test
+  // suite
```
Gudahtt, line 614: This comment was leftover from an earlier draft of this PR where some V3/V4 encoding was covered by the similarities test suite. That isn't the case anymore - only the similarities are covered there now. All behaviour for V3 and V4 are covered in their respective test suites. This comment can be deleted.

```suggestion
```

rekmarks (changes requested):
Finished a first pass. Overall looks good. I left some inline change requests.

I'm going to do one more pass to confirm that the table is fully covered, and consider whether the inputs are exhaustive or not. Then we should be done here!

src/index.test.ts
rekmarks: This was addressed.

src/index.test.ts
Gudahtt: I think they're equivalent functionally, but yes that might be easier to read. That sounds good to me.

src/index.test.ts
Gudahtt: Ah, right :sweat_smile: 
I was just following existing conventions (I think?). Probably better to use arrow functions, but maybe we can do this in one pass later on.

src/index.test.ts
Gudahtt: Done in https://github.com/MetaMask/eth-sig-util/pull/164/commits/6e313ad055739f8c536763b0fecc884b33f0af7b

PR #169, opened 2021-08-11
Add `TypedDataUtils.findTypeDependencies` tests

Gudahtt:
This function is now comprehensively tested.

PR #170, opened 2021-08-11
Add `TypedDataUtils.hashStruct` tests

Gudahtt:
This function should now be comprehensively tested. The tests mirror the `TypedDataUtils.encodeData` tests exactly. Any uses of
`hashStruct` in the older signature tests have been removed, as they are now redundant.

### DIFF
--- a/src/__snapshots__/index.test.ts.snap
+++ b/src/__snapshots__/index.test.ts.snap
@@ -304,6 +304,310 @@ exports[`TypedDataUtils.encodeData V4 should encode data with a recursive data t
 
 exports[`TypedDataUtils.encodeData V4 should encode data with custom type 1`] = `"a0cedeb2dc280ba39b857546d74f5549c3a1d7bdc2dd96bf881f76108e23dac2fc71e5fa27ff56c350aa531bc129ebdf613b772b6604664f5d8dbe21b85eb0c8cd54f074a4af31b4411ff6a60c9719dbd559c221c8ac3492d9d872b041d703d1b5aadf3154a261abdd9086fc627b61efca26ae5702701d05cd2305f7c52a2fc8"`;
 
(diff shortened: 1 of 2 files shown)


Gudahtt:
This depends upon #169

rekmarks (changes requested):
These tests do mirror the `encodeData` tests _exactly_, and I'm wondering why we're going through all the trouble. Given that the only operation of `hashStruct` is to call `ethUtil.keccak` with the return value of `encodeData`. Are we really trying to test `keccak`, here?

It seems to me that this entire PR could be reduced to a single test case:

```js
describe('hashStruct', () => {
  it('hashes the result from encodeData', () => {
    // set up some mocks
    expect(encodeData).toHaveBeenCalledOnce();
    expect(ethUtil.keccak).toHaveBeenCalledOnce();
  })
})
```

The test should probably be instrumented such that we're convinced that `keccak` is indeed called with the return value of `encodeData`, but you get the idea.

rekmarks:
I dismissed my previous review because I really meant it as a comment. I want to discuss the matter, but I won't ultimately block on it.

Gudahtt:
If we were to test just that `keccak` and `encodeData` were called, we'd be testing the internals of the function, not the behaviour of the function. If we swapped our `keccak` implementation out for another one that behaved slightly differently, or if we added operations before or after that call, or if the structure of the `encodeData` output changed, we'd miss all of those problems. Tests written like that fundamentally aren't useful - they don't do their job, and they make it more difficult to refactor.

The thoroughness here might not be useful though. I did consider testing just a few representative samples. I went with the whole test suite because it seemed just as easy, and I don't know how unlikely it might be for a broken `keccack` implementation to behave oddly for some `Buffer` inputs but not others.

rekmarks (approved):
LGTM!

PR #172, opened 2021-08-11
Add `TypedDataUtils.sanitizeData` tests

Gudahtt:
This function should now be comprehensively tested.

PR #173, opened 2021-08-11
Add `TypedDataUtils.eip712Hash` unit tests

Gudahtt:
Unit tests have been added for `TypedDataUtils.eip712Hash`. These tests don't exhaustively check all possible inputs because there are a lot, and they're just passed along to `hashStruct`. So the `hashStruct` tests should catch any input-specific changes. Outside of behaviour internal to `hashStruct`, all of the behaviour should be covered by these tests.

Any calls to `eip712Hash` in the older signature tests have been removed, as they are now redundant.

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -3064,6 +3064,9 @@ describe('TypedDataUtils.eip712Hash', function () {
   describe('V3', function () {
     it('should hash a minimal valid typed message', function () {
       const hash = sigUtil.TypedDataUtils.eip712Hash(
+        // This represents the most basic "typed message" that is valid according to our types.
+        // It's not a very useful message (it's totally empty), but it's complete according to the
+        // spec.
         {
           types: {
             EIP712Domain: [],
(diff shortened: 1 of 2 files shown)


Gudahtt:
This depends upon #172

rekmarks (commented):
My initial impression here is the same as https://github.com/MetaMask/eth-sig-util/pull/170#pullrequestreview-730218317, namely that these tests seem perhaps uselessly redundant. I'm not sure why it wouldn't suffice to check that it makes the expected calls to `sanitizeData`, `hashStruct`, and `ethUtil.keccak` given the possible branches inside the function, and rely on the `encodeData` tests for testing a wide range of inputs.

Gudahtt:
This has been rebased to resolve conflicts

Gudahtt:
I think these tests are mostly devoid of repeats actually! I only included a few sample inputs here. Most of it was focused on behaviour specific to this function, about how it encodes the type header and the data and concatenates them together, with the data being optional and omitted in some circumstances. And the peculiar range of expected inputs, since it will happily accept inputs that are contrary to the input types.

src/index.test.ts
```diff
@@ -3060,6 +3060,488 @@ describe('TypedDataUtils.sanitizeData', function () {
...
+        } as any,
+        'V3',
+      );
+
+      expect(minimalValidHash.toString('hex')).toBe(
+        extraPropertiesHash.toString('hex'),
+      );
+    });
+
+    it('should hash a typed message with a domain separator that uses all fields.', function () {
```
Gudahtt, line 3137: Nit: most test descriptions don't have a period
```suggestion
    it('should hash a typed message with a domain separator that uses all fields', function () {
```

rekmarks (approved):
LGTM!

PR #174, opened 2021-08-12
Improve`TypedDataUtils.eip712Hash` inline docs

Gudahtt:
The inline documentation for `TypedDataUtils.eip712Hash` had erroneously suggested that the function would sign the message. This is untrue, and has been corrected.


rekmarks (approved):
LGTM!

PR #175, opened 2021-08-12
Export functions directly rather than in object

Gudahtt:
Named exports are now exported directly on the line the function is declared, rather than as one large object. This is in-line with how we export things in all of our other TypeScript projects.

The `typedSignatureHash` internal function was renamed to avoid conflict with the exported `typedSignatureHash`, which differs only in the return type.


rekmarks (approved):
LGTM!

rekmarks (approved):
LGTM!

PR #176, opened 2021-08-12
Add `salt` to the EIP-712 `domain` type

Gudahtt:
The type for the `domain` property of an EIP-712 typed message has been updated to include `salt`, which is in the [EIP-712 specification](https://eips.ethereum.org/EIPS/eip-712#definition-of-encodedata).

The tests have been updated to include this property.

This is an optional property, so this is a non-breaking change.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -47,6 +47,7 @@ export interface TypedMessage<T extends MessageTypes> {
     version?: string;
     chainId?: number;
     verifyingContract?: string;
+    salt?: ArrayBuffer;
   };
   message: Record<string, unknown>;
 }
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -47,6 +47,7 @@ export interface TypedMessage<T extends MessageTypes> {
     version?: string;
     chainId?: number;
     verifyingContract?: string;
+    salt?: ArrayBuffer;
```
Gudahtt, line 50: I chose `ArrayBuffer` for the expected type because it seemed like the most reasonable way to communicate raw bytes, since it's the underlying data structure for typed arrays. This is the only cross-platform way to represent bytes I think :thinking:.

rekmarks (approved):
LGTM!

PR #178, opened 2021-08-12
Add docs and tests for `normalize`

Gudahtt:
Inline documentation and tests have been added for the `normalize` function. Only one test was added; the rest were pre-existing. They were grouped together and updated to match our typical testing conventions.

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -1712,29 +1712,38 @@ describe('concatSig', function () {
   });
 });
 
-it('normalize address lower cases', function () {
-  const initial = '0xA06599BD35921CfB5B71B4BE3869740385b0B306';
-  const result = sigUtil.normalize(initial);
-  expect(result).toBe(initial.toLowerCase());
-});
+describe('normalize', function () {
+  it('should normalize an address to lower case', function () {
+    const initial = '0xA06599BD35921CfB5B71B4BE3869740385b0B306';
+    const result = sigUtil.normalize(initial);
+    expect(result).toBe(initial.toLowerCase());
+  });
 
-it('normalize address adds hex prefix', function () {
-  const initial = 'A06599BD35921CfB5B71B4BE3869740385b0B306';
-  const result = sigUtil.normalize(initial);
-  expect(result).toBe(`0x${initial.toLowerCase()}`);
-});
+  it('should normalize address without a 0x prefix', function () {
+    const initial = 'A06599BD35921CfB5B71B4BE3869740385b0B306';
+    const result = sigUtil.normalize(initial);
+    expect(result).toBe(`0x${initial.toLowerCase()}`);
+  });
 
-it('normalize an integer converts to byte-pair hex', function () {
-  const initial = 1;
-  const result = sigUtil.normalize(initial);
-  expect(result).toBe('0x01');
-});
+  it('should normalize an integer to a byte-pair hex string', function () {
+    const initial = 1;
+    const result = sigUtil.normalize(initial);
+    expect(result).toBe('0x01');
+  });
 
-it('normalize an unsupported type throws', function () {
-  const initial = {};
-  expect(() => sigUtil.normalize(initial as any)).toThrow(
-    'eth-sig-util.normalize() requires hex string or integer input. received object:',
-  );
+  // TODO: Add validation to disallow negative integers.
+  it('should normalize a negative integer to 0x', function () {
+    const initial = -1;
+    const result = sigUtil.normalize(initial);
+    expect(result).toBe('0x');
+  });
+
+  it('should throw if given an object', function () {
+    const initial = {};
+    expect(() => sigUtil.normalize(initial as any)).toThrow(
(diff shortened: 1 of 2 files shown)


rekmarks (changes requested):
I imagine the purpose of the `object` test case is to prove to ourselves that inputs that are neither strings nor numbers will be rejected. That being the case, I think we should either note that, or just test a bunch of primitives as well, perhaps by iterating over an array of bad inputs.

src/index.ts
```diff
@@ -353,6 +353,11 @@ function concatSig(v: Buffer, r: Buffer, s: Buffer): string {
   return ethUtil.addHexPrefix(rStr.concat(sStr, vStr)).toString('hex');
 }
 
+/**
+ * Normalize the input to a 0x-prefixed hex string.
+ *
+ * @param input - The value to normalize
```
Gudahtt, line 359: The return value is missing here

Gudahtt: Resolved in https://github.com/MetaMask/eth-sig-util/pull/178/commits/038f8476225ed1fd0938375ab14a51d9b51f81d0

Gudahtt:
Good idea! I've added tests for all primitive types. I caught some unexpected behaviour for `undefined`, and `null`, so that was interesting.

src/index.test.ts
```diff
@@ -3023,29 +3023,71 @@ describe('concatSig', function () {
...
-    'eth-sig-util.normalize() requires hex string or integer input. received object:',
-  );
+  // TODO: Add validation to disallow negative integers.
+  it('should normalize a negative integer to 0x', function () {
+    const initial = -1;
+    const result = sigUtil.normalize(initial);
+    expect(result).toBe('0x');
+  });
+
+  // TODO: Add validation to disallow undefined.
```
rekmarks, line 3052:
```suggestion
  // TODO: Add validation to disallow null.
```

src/index.test.ts
```diff
@@ -3023,29 +3023,71 @@ describe('concatSig', function () {
...
+  it('should return undefined if given null', function () {
+    const initial = null;
+    expect(sigUtil.normalize(initial as any)).toBeUndefined();
+  });
+
+  // TODO: Add validation to disallow null.
```
rekmarks, line 3058:
```suggestion
  // TODO: Add validation to disallow undefined.
```

src/index.test.ts
```diff
@@ -3023,29 +3023,71 @@ describe('concatSig', function () {
...
+  });
+
+  it('should throw if given a bigint', function () {
+    const initial = BigInt(Number.MAX_SAFE_INTEGER);
+    expect(() => sigUtil.normalize(initial as any)).toThrow(
+      'eth-sig-util.normalize() requires hex string or integer input. received bigint: 9007199254740991',
+    );
+  });
+
+  it('should throw if given a sumbol', function () {
```
rekmarks, line 3085:
```suggestion
  it('should throw if given a symbol', function () {
```

src/index.test.ts
```diff
@@ -3023,29 +3023,71 @@ describe('concatSig', function () {
...
+  });
+
+  it('should throw if given a boolean', function () {
+    const initial = true;
+    expect(() => sigUtil.normalize(initial as any)).toThrow(
+      'eth-sig-util.normalize() requires hex string or integer input. received boolean: true',
+    );
+  });
+
+  it('should throw if given a bigint', function () {
```
rekmarks, line 3078:
```suggestion
  it('should throw if given a BigInt', function () {
```

Gudahtt: This one was lower-cased on purpose - apparently the primitive is called `bigint` but the object-wrapper is `BigInt. Kinda like `string` v.s. `String`

rekmarks (commented):
LGTM!

src/index.test.ts
Gudahtt: Whoops! Had those backwards I guess

rekmarks (approved):
No, it actually LGTM!

src/index.test.ts
Gudahtt: To reverse this: ```suggestion
  it('should throw if given a bigint', function () {
```

File: src/index.test.ts [hunk 6]
```diff
@@ -3023,29 +3023,71 @@ describe('concatSig', function () {
...
+  });
+
+  it('should throw if given a boolean', function () {
+    const initial = true;
+    expect(() => sigUtil.normalize(initial as any)).toThrow(
+      'eth-sig-util.normalize() requires hex string or integer input. received boolean: true',
+    );
+  });
+
+  it('should throw if given a BigInt', function () {
```
Gudahtt, line 3078: The primitive type is `bigint` apparently
```suggestion
  it('should throw if given a bigint', function () {
```

src/index.test.ts
Gudahtt: Second attempt: https://github.com/MetaMask/eth-sig-util/pull/178/files#r689795415

rekmarks (approved):

PR #179, opened 2021-08-15
Simplify findTypeDependencies implementation

rekmarks:
The implementation of this function left much to be desired. This PR refactors it to use a `Set` to store the dependencies and removes a completely extraneous inner loop.


src/index.ts
```diff
@@ -229,17 +231,17 @@ function encodeType(
...
     return results;
   }
-  results.push(primaryType);
+
+  results.add(primaryType);
+
   for (const field of types[primaryType]) {
-    for (const dep of findTypeDependencies(field.type, types, results)) {
-      !results.includes(dep) && results.push(dep);
-    }
```
rekmarks, line 242: `results` is the same object throughout the entire recursion!

Gudahtt: Haha, oh, good find.

src/index.ts
```diff
@@ -229,17 +231,17 @@ function encodeType(
 function findTypeDependencies(
   primaryType: string,
   types: Record<string, MessageTypeProperty[]>,
-  results: string[] = [],
-): string[] {
+  results: Set<string> = new Set(),
+): Set<string> {
```
Gudahtt, line 235: Unfortunately this function is part of this package's public API, which makes this a breaking change. We do have a few breaking changes already, but I was trying to minimize them for the sake of making the update process easier.

How important is this change? I think we could get rid of the loop but leave this as an array, and have one less breaking change.

rekmarks: Making it a Set is not _that_ important, and I see the point about breaking our API, however we at least don't use `findTypeDependencies` outside of this library: https://github.com/search?q=org%3AMetaMask+findTypeDependencies&type=code

rekmarks: I guess I could also return an array, that'd be non-breaking 🤔

rekmarks: I think we're fine either way! We could probably stop exporting this method entirely.

Gudahtt: This is one of our most popular libraries - it's not just used internally. It's relied upon by dozens of other projects, many of them fairly high profile (e.g. ganache, hardhat, Ledger, matic, burner-wallet, etc.)

I wasn't able to find any usages of this particular function when I searched, but I did find that other functions like `encodeData` that I had expected to be unused were relied upon by third parties. So who knows.

rekmarks: Alright, we'll keep the function, but since we're shipping a major version anyway, I'd prefer that we just break the API of this function, too. Finding the unique elements from a (conceptual) set with repetitions is an ideal use case for a `Set`. This library hasn't had a facelift in years, I don't think our consumers will quit on us.

Gudahtt: OK, if you feel it's important we can break it. We can include migration instructions with the changelog entry.

rekmarks:
We're up to date again (for now).

Gudahtt (approved):
LGTM!

PR #180, opened 2021-08-15
Improve padWithZeroes implementation and add tests

rekmarks:
The `padWithZeroes` implementation was inefficient, but most of all, ugly. This PR makes it more efficient, adds some basic input validation, and unit tests.


src/index.ts
```diff
@@ -686,12 +686,32 @@ function getPublicKeyFor<T extends MessageTypes>(
   return recoverPublicKey(msgHash, msgParams.sig);
 }
 
-function padWithZeroes(number: string, length: number): string {
-  let myString = `${number}`;
-  while (myString.length < length) {
-    myString = `0${myString}`;
```
rekmarks, line 692: This computes a new string and stores it in memory `length - number.length` times.

src/index.ts
```diff
@@ -686,12 +686,32 @@ function getPublicKeyFor<T extends MessageTypes>(
...
+ *
+ * If the input string is "0x"-prefixed or not a hex string, an error will be
+ * thrown.
+ *
+ * @param hexString The hexadecimal string to pad with zeroes.
+ * @param targetLength The target length of the hexadecimal string.
+ * @returns The input string front-padded with zeroes, or the original string
+ * if it was already greater than or equal to to the target length.
+ */
+export function padWithZeroes(hexString: string, targetLength: number): string {
```
Gudahtt, line 702: Gotta say, I'm not too keen on adding more stuff to this API :thinking: We've already exposed an annoying amount of what would otherwise be internal functions.

I think it would be fine to test this via `concatSig`, remove the tests, and keep this private.

Gudahtt: Either that or delete this altogether and use `String.prototype.padStart`

rekmarks: TIL `padStart` is a thing! I'm definitely going to use that.

rekmarks:
@Gudahtt I refactored it to use `padStart` as you suggested, and moved the file to a new file, `utils.ts` that has its own tests.

src/utils.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Pads the front of the given hex string with zeroes until it reaches the
+ * target length. If the input string is already longer than or equal to the
+ * target length, it is returned unmodified.
+ *
+ * If the input string is "0x"-prefixed or not a hex string, an error will be
+ * thrown.
+ *
+ * @param hexString The hexadecimal string to pad with zeroes.
+ * @param targetLength The target length of the hexadecimal string.
```
Gudahtt, line 10: Nit: We typically separate parameter descriptions from the variable name with a ` - `, as is done in the JSDoc docs (though it seems to be optional according to JSDoc IIRC)

```suggestion
 * @param hexString - The hexadecimal string to pad with zeroes.
 * @param targetLength - The target length of the hexadecimal string.
```

src/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+ * @returns The input string front-padded with zeroes, or the original string
+ * if it was already greater than or equal to to the target length.
+ */
+export function padWithZeroes(hexString: string, targetLength: number): string {
+  if (hexString !== '' && !/^[a-f0-9]+$/iu.test(hexString)) {
+    throw new Error(
+      `Expected an unprefixed hex string. Received: ${hexString}`,
```
Gudahtt, line 17: The old function didn't include this check. It would coerce anything it was given into a string. I'm pretty confident it was only ever called with a hex string though, because in the one place it's called, it's given the result of a `Buffer.prototype.toString('hex')` call. So I'd guess this doesn't change behaviour.

src/utils.test.ts
```diff
@@ -0,0 +1,27 @@
...
+    const input = 'abc';
+    expect(padWithZeroes(input, 5)).toStrictEqual(`00${input}`);
+  });
+
+  it('pads an empty string', function () {
+    const input = '';
+    expect(padWithZeroes(input, 4)).toStrictEqual(`0000`);
+  });
+
+  it('returns a string longer than or equal to the target length without modifying it', function () {
```
Gudahtt, line 14: Nit: might be nice to test the "longer than" case, which isn't covered here.

Gudahtt (approved):
LGTM! Left a couple of minor non-blocking suggestions

src/utils.ts
rekmarks: Yeah, I think it is optional, but you are correct that it is the convention!

src/utils.ts
rekmarks: Yeah, my thinking was exactly that it's non-breaking for current usage, and we won't accidentally shoot ourselves in the foot in the future.

rekmarks:
Okay, I added a couple more test cases, including the one you suggested. I also added input validation for the target length.

Gudahtt (approved):
LGTM!

PR #181, opened 2021-08-17
Add inline docs for all public functions

Gudahtt:
All public functions now have inline documentation. Some of these descriptions are rather sparse, but it's a start. We can improve them later.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -436,7 +436,7 @@ export function recoverPersonalSignature<T extends MessageTypes>(
 }
 
 /**
- * Recover the public key of the account used to  create the given Ethereum signature. The message
+ * Recover the public key of the account used to create the given Ethereum signature. The message
  * must have been signed using the `personalSign` function, or an equivalent function.
  *
  * @param msgParams - The message parameters, which includes both the message and the signature.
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -396,6 +396,16 @@ export function normalize(input: number | string): string {
   return ethUtil.addHexPrefix(input.toLowerCase());
 }
 
+/**
+ * Create an Ethereum-specific signature for a message.
+ *
+ * This function is equivalent to the `eth_sign` Ethereum JSON-RPC method as specified in EIP-1417,
+ * as well as the MetaMask provider `personal_sign` method.
```
rekmarks, line 403: I am on a crusade to make "provider" a proscribed term, at least until we can agree on what it means!

```suggestion
 * as well as the MetaMask's `personal_sign` method.
```

src/index.ts
```diff
@@ -416,18 +435,44 @@ export function recoverPersonalSignature<T extends MessageTypes>(
   return senderHex;
 }
 
+/**
+ * Recover the public key of the account used to  create the given Ethereum signature. The message
```
rekmarks, line 439:
```suggestion
 * Recover the public key of the account used to create the given Ethereum signature. The message
```

rekmarks (changes requested):
Much-needed!

src/index.ts
Gudahtt: Sure! This still reads fine without that term.

rekmarks (approved):
LGTM!

PR #182, opened 2021-08-17
Add `personalSign` tests

Gudahtt:
The `personalSign` function is now more thoroughly tested. The existing tests were grouped together and made more concise, and a few extra tests were added to ensure that the exact signature is stable for the test input.

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -3714,40 +3714,78 @@ describe('normalize', function () {
   });
 });
 
-it('personalSign and recover', function () {
-  const address = '0x29c76e6ad8f28bb1004902578fb108c507be341b';
-  const privKeyHex =
-    '4af1bceebf7f3634ec3cff8a2c38e51178d5d4ce585c52d6043e5e2cc3418bb0';
-  const privKey = Buffer.from(privKeyHex, 'hex');
-  const message = 'Hello, world!';
-  const msgParams: sigUtil.MsgParams<string> = { data: message };
-
-  const signed = sigUtil.personalSign(privKey, msgParams);
-  msgParams.sig = signed;
-  const recovered = sigUtil.recoverPersonalSignature(
-    msgParams as sigUtil.SignedMsgParams<string>,
+describe('personalSign', function () {
+  const privateKey = Buffer.from(
+    '4af1bceebf7f3634ec3cff8a2c38e51178d5d4ce585c52d6043e5e2cc3418bb0',
+    'hex',
   );
+  const helloWorldSignature =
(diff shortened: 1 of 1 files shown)


src/index.test.ts
```diff
@@ -3714,40 +3714,78 @@ describe('normalize', function () {
...
-  const signed = sigUtil.personalSign(privKey, msgParams);
-  msgParams.sig = signed;
-  const recovered = sigUtil.recoverPersonalSignature(
-    msgParams as sigUtil.SignedMsgParams<string>,
+describe('personalSign', function () {
+  const privateKey = Buffer.from(
+    '4af1bceebf7f3634ec3cff8a2c38e51178d5d4ce585c52d6043e5e2cc3418bb0',
+    'hex',
   );
+  const helloWorldSignature =
```
rekmarks, line 3722: Isn't this a good place for a snapshot? I'm genuinely curious – I've never used snapshots myself.

rekmarks (changes requested):

src/index.test.ts
Gudahtt: It would be, except that this is the input for the recovery tests.

Generally snapshots are great for cases where we want to ensure the output doesn't change, and where the output isn't readable. Even then, inline snapshots should be preferred for shorter outputs, so that it's easier to see what's going on when they change. I try to use them as a last resort, but I did find them very useful for opaque signatures.

rekmarks: Ah, I didn't pay close enough attention to where it was used elsewhere. Thanks!

src/index.test.ts
```diff
@@ -3714,40 +3714,78 @@ describe('normalize', function () {
...
+      data: helloWorldMessage,
+    });
+
+    expect(
+      sigUtil.extractPublicKey({
+        data: helloWorldMessage,
+        sig: signature,
+      }),
+    ).toBe(publicKey);
+  });
```
rekmarks, line 3788: The purpose of these tests appears to be to convince ourselves that we can create new, valid signatures and extract the public key and address from them. That being the case, wouldn't it be better to use some other data to sign that the "hello world" message? Otherwise I don't think they're telling us anything we don't already now.

rekmarks (changes requested):

src/index.test.ts
rekmarks: We talked this over, it's fine.

rekmarks (approved):
LGTM!

src/index.test.ts
Gudahtt: The other two tests ensure that we generate the same signature we did before, and that we can recover the same signature we did before. These are important because signatures might be stored by third parties, so any change to the format would be breaking.

But this test just ensures the public key is recoverable, irrespective of whether the signature has changed or not. This test would be useful if we ever intentionally changed the signature output for some reason. Or I guess it might be useful in pinpointing what went wrong if we accidentally changed it.

PR #183, opened 2021-08-17
Add `.prettierrc.js`

Gudahtt:
This file was accidentally omitted when updating the linter. As a result `prettier` and `ESLint` have been clashing over the formatting, which slows down the linter. They are now aligned, and the linter is faster.


rekmarks (approved):
A crucial find!

PR #184, opened 2021-08-17
Fix `eip712Hash` tests that included data

Gudahtt:
The `eip712Hash` tests that included data had the wrong `primaryType`, so the data wasn't actually included in the resulting hash. This was an edge case that had not been considered, so another test was added to capture it.


rekmarks (approved):
LGTM!

PR #185, opened 2021-08-17
Add `signTypedData` and `recoverTypedSignature` tests

Gudahtt:
The `signTypedData` and `recoverTypedSignature` functions have been thoroughly tested, using all three versions. Older redundant tests have been removed.


src/index.test.ts
```diff
@@ -3788,176 +3791,2407 @@ describe('personalSign', function () {
...
+      ],
+      int8: [0, '0', '0x0', 255, -255],
+      int256: [0, '0', '0x0', Number.MAX_SAFE_INTEGER, Number.MIN_SAFE_INTEGER],
+      uint8: [0, '0', '0x0', 255, -255],
+      uint256: [
+        0,
+        '0',
+        '0x0',
+        Number.MAX_SAFE_INTEGER,
+        Number.MIN_SAFE_INTEGER,
```
Gudahtt, line 3840: I had to create a separate set of example data for V1 because it differed from the V3/V4 behaviour a fair bit.

Here's the first highlight - negative unsigned data! Somehow it doesn't throw.

src/index.test.ts
```diff
@@ -3788,176 +3791,2407 @@ describe('personalSign', function () {
...
-    {
-      type: 'string',
-      name: 'message',
-      value: 'Hi, Alice!',
-    },
-  ];
+    const signTypedDataV1ErrorExamples = {
+      string: [
+        {
+          input: 10,
```
Gudahtt, line 3852: Second highlight - it doesn't accept numbers as string data. Makes sense, but it's unfortunate that V3 and V4 do.

src/index.test.ts
```diff
@@ -3788,176 +3791,2407 @@ describe('personalSign', function () {
...
+      string: [
+        'Hello!',
+        '0xbBbBBBBbbBBBbbbBbbBbbbbBBbBbbbbBbBbbBBbB',
+        '0xabcd',
+        '😁',
+      ],
+      // atomic types supported by EIP-712:
+      address: [
+        '0xbBbBBBBbbBBBbbbBbbBbbbbBBbBbbbbBbBbbBBbB',
+        '0xbBbBBBBbbBBBbbbBbbBbbbbBBbBbbbbBbBbbBBbBbBbBBBBbbBBBbbbBbbBbbbbBBbBbbbbBbBbbBBbB', // no apparent maximum length
```
Gudahtt, line 3808: Third highlight - infinite-length addresses!

src/index.test.ts
```diff
@@ -3788,176 +3791,2407 @@ describe('personalSign', function () {
...
+      string: [
+        {
+          input: 10,
+          errorMessage:
+            'The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received type number (10)',
+        },
+      ],
+      address: [
+        {
+          input: 'bBbBBBBbbBBBbbbBbbBbbbbBBbBbbbbBbBbbBBbB',
```
Gudahtt, line 3859: Fourth highlight - apparently unprefixed addresses are not allowed with V1. They are accepted by V3/V4.

src/index.test.ts
```diff
@@ -3788,176 +3791,2407 @@ describe('personalSign', function () {
...
-    },
-    {
-      type: 'uint8',
-      name: 'value',
-      value: 10,
-    },
-  ];
+          if (type === 'bytes') {
+            _it(
+              `should fail to sign array of all ${type} example data`,
```
Gudahtt, line 3946: Fifth highlight - V1 supports arrays of any type except `bytes`. Apparently an array of `bytes` is unacceptable.

Gudahtt:
This has just been rebased to resolve conflicts.

rekmarks (changes requested):
This looks good! Excited to close this chapter.

 I reviewed the `V1` and `recoverTypedSignature` tests by hand, and relied extensively on `diff` to compare most of the `V3` and `V4` tests (which are indeed mirror images of their `encodeData` and `eip712Hash` counterparts).

I've left one inline question / comment and requested some additional changes in #191. Otherwise, my only general request is that we split up this humongous file in a follow-up 😅

src/index.test.ts
Gudahtt: Anything can be in an array though. From the page you just linked:

>Array elements can be of any type, including mapping or struct.

So this is an array of `bytes`. An array of an array of bytes really.

PR #186, opened 2021-08-17
Fix `typedSignatureHash` documentation

Gudahtt:
The inline docs erroneously described this function as returning a type hash for "V1" of `signTypedData`. But "V1" didn't include a type hash. Rather, this function returns the hash of the entire message.


rekmarks (approved):
LGTM!

PR #187, opened 2021-08-18
Add `typedSignatureHash` tests

Gudahtt:
There were a surprising number of `typedSignatureHash` tests already, but they have been expanded and grouped together.

The same dataset used for the `signTypedData_v1` tests was moved to the module scope so that it could be used for these tests as well.


Gudahtt:
This depends upon #185

rekmarks (approved):
LGTM!

PR #189, opened 2021-08-19
Add solidity types to schema

Gudahtt:
The solidity types supported by EIP-712 have been added to the typed message JSON schema.

This was originally proposed in #31; see there for more details.


rekmarks (approved):
LGTM!

PR #192, opened 2021-08-26
Fix `concatSig` parameter type

Gudahtt:
The `v` parameter passed into the `concatSig` function internally was always a number, but it was incorrectly typed and treated as a Buffer. The function used to convert the input from a Buffer into an integer would convert the integer to a Buffer, then back into an integer.

Now `concatSig` accepts `v` as either an integer or a Buffer. If given an integer, it skips the redundant conversions it had done before.


Gudahtt:
I'm going to close this in favour of instead disallowing `number` inputs, because apparently `ethereumjs-util@7` uses `Buffer` for `v` anyway.

Gudahtt:
This has been replaced by #194

PR #194, opened 2021-08-30
Fix `concatSig` parameter and handling of return value

Gudahtt:
The `v` parameter of the `concatSig` function was being passed in as a `number`, but it was supposed to be a `Buffer`. It is now converted to `Buffer` before being passed in. This should not be a functional change.

The return value was being treated as a `Buffer`, but it was a hex-prefixed string. The useless Buffer-to-hex-string conversion step has been removed.


rekmarks (approved):
LGTM!

PR #195, opened 2021-08-31
Update `ethereumjs-util` from v5 to v6

Gudahtt:
The `ethereumjs-util` library has been updated. Some methods which were just re-exported from `ethjs-util` are now used directly because v6 of `ethereumjs-util` omits those functions from its type definitions.

The v6 version of `toBuffer` no longer accepts non-hex-prefixed strings, so a private `legacyToBuffer` function has been created to convert values in the same manner as v5 used to. Non-hex-prefixed strings are converted using `Buffer.from`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,8 @@
   },
   "dependencies": {
     "ethereumjs-abi": "^0.6.8",
-    "ethereumjs-util": "^5.2.1",
+    "ethereumjs-util": "^6.2.1",
+    "ethjs-util": "^0.1.6",
     "tweetnacl": "^1.0.3",
     "tweetnacl-util": "^0.15.1"
   },
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,6 +2,7 @@ import * as ethUtil from 'ethereumjs-util';
 import * as ethAbi from 'ethereumjs-abi';
 import * as nacl from 'tweetnacl';
 import * as naclUtil from 'tweetnacl-util';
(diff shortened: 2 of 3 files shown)


Gudahtt:
This depends upon #194

Gudahtt:
For reference, [here is the old version of `toBuffer`](https://github.com/ethereumjs/ethereumjs-util/blob/v5.2.1/index.js#L159), and [here is the new version of `toBuffer`](https://github.com/ethereumjs/ethereumjs-util/blob/v6.2.1/src/bytes.ts#L70) that this PR is updating to.

Also note that `ethereumjs-util` has since been updated to v7, but I'm leaving that for a future PR because using v6 will make replacing `ethereumjs-abi` easier because `ethereumjs-abi`  uses `ethereumjs-util@6` internally.

package.json
```diff
@@ -37,7 +37,8 @@
   },
   "dependencies": {
     "ethereumjs-abi": "^0.6.8",
-    "ethereumjs-util": "^5.2.1",
+    "ethereumjs-util": "^6.2.1",
+    "ethjs-util": "^0.1.6",
```
rekmarks, line 41: Was this used internally by `ethereumjs-util@5` or something?

rekmarks (approved):
LGTM!

Koagonzalo11:
[https://etherscan.io/0x0df1e8465f3bf68bc20857208472dcbec7092472/](url)

Koagonzalo11:
[launch.json](https://github.com/MetaMask/eth-sig-util/files/14597252/launch.json)

PR #197, opened 2021-09-02
Consolidate the data encoding for `signTypedData` V3 and V4

Gudahtt:
The encoding logic for `signTypedData` was nearly identical between V3 and V4, but each version was a completely separate implementation. This complicated my efforts to replace `ethereumjs-abi`. The logic for each version has been consolidated into one function.

The main differences are:
* `undefined` values are skipped in V3.
* `undefined` values are encoded as zero on V4 for custom types, but they throw an error for non-custom types.
* `null` types are encoded as zero on V4 for custom types, but are the same between versions for non-custom types.
* V3 does not support arrays.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -93,6 +93,72 @@ function getSolidityTypes() {
 }
 
 /**
+ * Encode a single field.
+ *
+ * @param types - All type definitions.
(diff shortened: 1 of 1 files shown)


rekmarks (approved):
LGTM!

PR #198, opened 2021-09-09
Simplify function type signatures

Gudahtt:
The type signatures of many public functions were complex and misleading. This was primarily due to the widespread use of the types `TypedData` and `TypedMessage<T>`, which where inappropriate in most of the places they were used.

The `TypedMessage<T>` type is used by the `signTypedData` function when the version is 'V3' or 'V4'. All references to this type outside of V3 and V4 of `signTypedData` have been removed. This also resulted in the `T` generic parameter being removed from many functions.

The `TypedData` type didn't represent the actual type signature of any function, yet it was used to represent the input data for most functions. In most cases we had no specific expectations for the input type, so `unknown` is now used instead. In the case of `V1` of `signTypedData`, the expected type was always `EIP712TypedData[]`. The `TypedData` type has been reduced to just that one case where it was useful and accurate.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,8 +6,18 @@ import { intToHex, isHexString, stripHexPrefix } from 'ethjs-util';
 
 import { padWithZeroes } from './utils';
 
-export type TypedData = EIP712TypedData[];
+/**
+ * This is the message format used for `V1` of `signTypedData`.
+ */
+export type TypedDataV1 = EIP712TypedData[];
 
+/**
+ * This represents a single field in a `V1` `signTypedData` message.
+ *
+ * @property name - The name of the field.
+ * @property type - The type of a field (must be a supported Solidity type).
+ * @property value - The value of the field.
+ */
 interface EIP712TypedData {
   name: string;
   type: string;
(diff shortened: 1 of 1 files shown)


Gudahtt:
The Node.js v16 failure is unrelated (it also fails on `main`). It is fixed here: https://github.com/MetaMask/eth-sig-util/pull/199

src/index.ts
```diff
@@ -6,7 +6,7 @@ import { intToHex, isHexString, stripHexPrefix } from 'ethjs-util';
 
 import { padWithZeroes } from './utils';
 
-export type TypedData = string | EIP712TypedData | EIP712TypedData[];
+export type TypedData = EIP712TypedData[];
```
rekmarks, line 9: If `TypedData` is only used in / for V1, can we change its name to `V1TypedData` or something similar?

rekmarks (changes requested):
In addition to the inline request, I think we should add docstrings to `TypedData` and `TypedMessage` noting what (i.e. which versions) they're for.

src/index.ts
Gudahtt: We could :thinking: I had left it as-is to minimize breaking changes, but I guess we're past that mattering now.

Gudahtt: Updated! It's now called `TypedDataV1`

Gudahtt:
Good idea! The latest commit adds docs for both of those types.

rekmarks (approved):
LGTM!

PR #200, opened 2021-09-09
Remove `signTypedData` version `V2`

Gudahtt:
This version was never documented or supported. In effect if anyone was using this, they were using `V3`.


rekmarks (approved):
LGTM!

PR #201, opened 2021-09-09
Add `signTypedData` version validation

Gudahtt:
The `Version` parameter used by `signTypedData`-related functions has been converted to an enum, and all `version` parameters are now validated against this enum.

The `TypedDataUtils` functions were written specifically for `V3` and `V4`, so `V1` is disallowed. If anyone was using these functions with `V1`, they were effectively using the `V3` versions.

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -1,6 +1,8 @@
 import * as ethUtil from 'ethereumjs-util';
 import * as sigUtil from '.';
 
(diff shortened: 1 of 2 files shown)


Gudahtt:
This depends upon #200

rekmarks (approved):
This should have been done from the start. LGTM!

PR #202, opened 2021-09-09
Reorganize old tests

Gudahtt:
All tests are now inside of `describe` blocks that indicate the general category of the group of tests. The encryption-related tests have been grouped together, and the last remaining old `personalSign` tests have been moved into the `personalSign` describe block.

The old `signatureTest` function used to run the `personalSign` test cases was replaced with a `for` loop that better matches the style of the other tests.

At a later point we should test the encryption functions independently from each other, but for now I'm going to leave them as-is. They also need their coverage improved.

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -4113,6 +4113,72 @@ describe('personalSign', function () {
       }),
     ).toBe(publicKey);
   });
+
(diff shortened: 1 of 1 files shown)


rekmarks (approved):
LGTM!

PR #203, opened 2021-09-09
Remove redundant dependencies from lockfile

Gudahtt:
A few redundant dependencies have been removed from the lockfile, most notably including some older cryptographic dependencies.


rekmarks (approved):
LGTM!

PR #204, opened 2021-09-10
Use options bag parameters rather than MsgParams type

Gudahtt:
The signing and encryption functions all accepted a `MsgParams` parameter that included just a single property (`data`, the data to be signed). This pointless wrapper object around this single parameter complicated the type signatures of these functions.

Instead, these functions now accept all parameters as one "options bag" parameter, and the `MsgParams` type has been removed. This makes the functions easier to use, as now it's not possible to enter the parameters in the wrong order.

The related `SignedMsgParams` type has also been replaced in the same manner, for the same reasons.

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -4145,7 +4145,7 @@ describe('personalSign', function () {
   const helloWorldMessage = 'Hello, world!';
 
   it('should sign a message', function () {
-    expect(sigUtil.personalSign(privateKey, { data: helloWorldMessage })).toBe(
+    expect(sigUtil.personalSign({ privateKey, data: helloWorldMessage })).toBe(
       helloWorldSignature,
     );
   });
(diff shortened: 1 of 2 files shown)


Gudahtt:
The `MsgParams` options didn't make a lot of sense, so the options were to use the "options bag" pattern or positional arguments. These are functions where getting the arguments in the wrong order could be pretty bad, and the team seems to like the "options bag" pattern and has introduced it in many places lately, so I went with that.

shanejonas:
Looks good. thats a lot of test fixin 😅 

would be nice to get rid of the `_expect` and `_it` and just disable that lint rule, but not with to this PR.

shanejonas (approved):

Gudahtt:
>would be nice to get rid of the _expect and _it and just disable that lint rule, but not with to this PR.

Yep! It's _awful_. I'll have a second look for a workaround then remove it.

PR #205, opened 2021-09-10
Add validation to check that parameters aren't nullish

Gudahtt:
Validation has been added to the signing and encryption functions to ensure that the parameters are not nullish. This should help with migrating from the last major version, as it would be easy to accidentally use the wrong parameter name when migrating.

A nullish check was used rather than a falsy check because there are falsy values that are valid for some parameters (e.g. the data being signed can be `0`). `null` and `undefined` are never valid in these cases though.

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -4272,6 +4272,122 @@ describe('personalSign', function () {
       _expect(recovered).toBe(addressHex);
     });
   }
+
+  describe('validation', function () {
+    describe('personalSign', function () {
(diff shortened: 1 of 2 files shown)


Gudahtt:
Further validation was intentionally delayed until later because that gets awfully complicated if we're also trying to avoid breaking changes. There are a lot of "invalid" parameters that will still produce valid signatures.

shanejonas (approved):

PR #212, opened 2021-09-20
Fix missing parameter descriptions in docs

Gudahtt:
The parameter descriptions were missing for any destructured parameters. Apparently TypeDoc doesn't support showing descriptions for destructured parameters ([see here for more details](https://github.com/TypeStrong/typedoc/issues/1703)).

I have forked TypeDoc to fix this problem, and published this fork as `@gudahtt/typedoc@0.23.0`. We can use this fork until this has been resolved upstream.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "tweetnacl-util": "^0.15.1"
   },
   "devDependencies": {
+    "@gudahtt/typedoc": "^0.23.0",
     "@lavamoat/allow-scripts": "^1.0.6",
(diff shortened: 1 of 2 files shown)


ryanml (approved):
nit: would it be more appropriate to fork this to the org? I'm fine with it either way

Gudahtt:
I could! I was hoping this would be a short-term thing. I can migrate it to the MetaMask org if I have trouble landing this upstream.

ryanml:
sounds good! lgtm

PR #218, opened 2021-09-21
Rename `signTypedData`-related types.

Gudahtt:
The type `EIP712TypedData` has been renamed to `TypedDataV1Field` so that it does not get confused for a valid `signTypedData` message or field. This type is only used for `V1` of `signTypedData`, which was based on an early draft version of EIP-712 and doesn't reflect the state of that EIP today at all.

The type `Version` has been renamed to `SignTypedDataVersion`. This ensures it doesn't get confused for an encryption version, for example. It's directly exported from this package, so it might not be clear at first that it's related to `signTypedData`.

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -10,7 +10,7 @@ Array [
   "personalSign",
   "recoverPersonalSignature",
   "extractPublicKey",
-  "Version",
+  "SignTypedDataVersion",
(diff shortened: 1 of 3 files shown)


rekmarks (approved):
LGTM!

01SMD (commented):
.

PR #219, opened 2021-09-21
Prevent tests from being compiled

Gudahtt:
A mistake in `tsconfig.json` resulted in tests being compiled when `yarn build` was run. The tests should now be correctly ignored when building.


rekmarks (approved):
LGTM!

PR #220, opened 2021-09-23
Update ESLint config from v7 to v9

Gudahtt:
This update lets us remove all of the JSDoc rules, as they are now included in our base ESLint config.


ryanml (approved):
LGTM!

PR #243, opened 2022-04-19
Fix typed message schema mistake

### LINKED ISSUE #242: TYPED_MESSAGE_SCHEMA does not support reference types
The exported JSON schema `TYPED_MESSAGE_SCHEMA` was updated in v4.0.0 to list all Solidity types that are valid in EIP-712 typed messages. Unfortunately, this accidentally disallowed the use of _reference types_ (i.e. using custom types within other types). This was unintended.

To give a specific example, this message was valid according to the `TYPED_MESSAGE_SCHEMA` schema prior to v4.0.0, but is invalid after v4.0.0:

```
{
  domain: {},
  message: {},
  primaryType: 'Message',
  types: {
    CustomValue: [{ name: 'value', type: 'string' }],
    EIP712Domain: [],
    Message: [{ name: 'data', type: 'CustomValue' }],
  },
}
```

The schema should be updated to allow for the use of reference types within reference types.

Gudahtt:
The exported JSON schema `TYPED_MESSAGE_SCHEMA` was changed in v4 to specify all supported Solidity types, but this accidentally prevented the use of reference types.

This change was undone, so that the schema no longer validates that the types used are valid. I could not find a way to preserve this validation without disallowing reference types. JSON schema did not make this possible.

Tests were added for the typed message schema to ensure this doesn't happen again.

Fixes #242

### DIFF
--- a/src/sign-typed-data.test.ts
+++ b/src/sign-typed-data.test.ts
@@ -1,16 +1,231 @@
 import * as ethUtil from 'ethereumjs-util';
+import Ajv from 'ajv';
 import {
   recoverTypedSignature,
   signTypedData,
   TypedDataUtils,
   typedSignatureHash,
   SignTypedDataVersion,
+  TYPED_MESSAGE_SCHEMA,
 } from './sign-typed-data';
 
 const privateKey = Buffer.from(
   '4af1bceebf7f3634ec3cff8a2c38e51178d5d4ce585c52d6043e5e2cc3418bb0',
   'hex',
 );
+
+/**
+ * Get a list of all Solidity types supported by EIP-712.
+ *
+ * @returns A list of all supported Solidity types.
+ */
+function getEip712SolidityTypes() {
+  const types = ['bool', 'address', 'string', 'bytes'];
+  const ints = Array.from(new Array(32)).map(
+    (_, index) => `int${(index + 1) * 8}`,
+  );
+  const uints = Array.from(new Array(32)).map(
+    (_, index) => `uint${(index + 1) * 8}`,
+  );
+  const bytes = Array.from(new Array(32)).map(
+    (_, index) => `bytes${index + 1}`,
+  );
+
+  return [...types, ...ints, ...uints, ...bytes];
+}
+
+const eip712SolidityTypes = getEip712SolidityTypes();
+
+describe('TYPED_MESSAGE_SCHEMA', () => {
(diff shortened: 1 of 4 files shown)


src/sign-typed-data.test.ts
```diff
@@ -1,17 +1,232 @@
...
+
+  return [...types, ...ints, ...uints, ...bytes];
+}
+
+const eip712SolidityTypes = getEip712SolidityTypes();
+
+describe('TYPED_MESSAGE_SCHEMA', () => {
+  it('should match valid typed message', () => {
+    const ajv = new Ajv();
+    const validate = ajv.compile(TYPED_MESSAGE_SCHEMA);
```
mcmire, line 42: [nit] What do you think about extracting this `validate` function since it's the same for every test?

src/sign-typed-data.test.ts
```diff
@@ -1,17 +1,232 @@
...
+        EIP712Domain: [],
+        Message: [],
+      },
+    };
+
+    expect(validate(typedMessage)).toBe(true);
+  });
+
+  for (const solidityType of eip712SolidityTypes) {
+    // eslint-disable-next-line no-loop-func
```
mcmire, line 72: [nit] If you use `.forEach` instead of a `for` loop, does it remove the need for this override?

mcmire (approved):
That's very interesting that JSON Schema didn't allow you to do what you wanted to do.

A couple of nits below but looks good to me regardless.

src/sign-typed-data.test.ts
Gudahtt: It does, but now that you've brought this up, I kinda want to get to the bottom of this. This ought to work. Something is messed up with the ESLint globals configuration. I'll look into this in a subsequent PR.

src/sign-typed-data.test.ts
Gudahtt: Good idea! Done in b5f0d877f72fb36fd41d64eec7a8cb248aba33f5

src/sign-typed-data.test.ts
Gudahtt: Oh my goodness. I found the problem. The override for the test ESLint config is wrong, the file glob uses the `js` file extension rather than `ts`. There are a bunch of violations, I'll fix this separately.

mcmire (approved):

PR #244, opened 2022-04-20
Add failing test for BN overflow issue

yvettep321:
https://github.com/MetaMask/metamask-extension/issues/7881

### DIFF
--- a/test/index.ts
+++ b/test/index.ts
@@ -578,7 +578,7 @@ test('signedTypeData_v4', (t) => {
     domain: {
         name: 'Ether Mail',
         version: '1',
-        chainId: 1,
+        chainId: 9007199254740992,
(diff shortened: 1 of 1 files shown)


test/index.ts
```diff
@@ -578,7 +578,7 @@ test('signedTypeData_v4', (t) => {
     domain: {
         name: 'Ether Mail',
         version: '1',
-        chainId: 1,
+        chainId: 9007199254740992,
```
mcmire, line 581: Instead of changing this test can we add a new test? Otherwise the `try` part of the `try`/`catch` will never get hit.

Gudahtt:
It seems like this was a mistake? If not, please explain your intent with creating this PR @yvettep321.

This commit is an old branch Finlay tested to create a repro for an extension bug.

PR #245, opened 2022-04-20
Fix lint rules for tests

Gudahtt:
The lint configuration for tests was wrong (the file glob had the wrong file extension), so none of the test lint rules or globals were being set correctly. This has been fixed, and all rule violations have been fixed.

There were a number of workarounds used in the tests for the globals `expect` and `it` not being recognized by the linter. These workarounds have now been removed as well, since globals are now correctly set.


Gudahtt:
This depends upon #243

.eslintrc.js
```diff
@@ -7,11 +7,10 @@ module.exports = {
       extends: ['@metamask/eslint-config-typescript'],
     },
     {
-      files: ['*.test.js'],
+      files: ['*.test.ts'],
       extends: ['@metamask/eslint-config-jest'],
     },
   ],
-  plugins: ['jsdoc'],
```
Gudahtt, line 14: This plugin is already added in our general shared config, so this was redundant.

mcmire (approved):

PR #249, opened 2022-05-13
Update to latest `typedoc` version

Gudahtt:
A fork of `typedoc` was being used because it included [an enhancement to how destructured parameters are interpreted](https://github.com/TypeStrong/typedoc/pull/1704). This enhancement has since been merged and published as [v0.22.5](https://github.com/TypeStrong/typedoc/releases/tag/v0.22.5), so we don't need to use the fork anymore.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,6 @@
     "tweetnacl-util": "^0.15.1"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


mcmire (approved):
Nice!

PR #253, opened 2022-06-02
Decryption using shared secret support for hardware wallet

luall0:
Hello there !

I've implemented decryptWithSharedSecret function that allow to decrypt encrypted data with an shared secret (that will allow to compute ephemeral private key) rather than the private key itself.
This will be really helpfull for hardware wallet operation (you ask the hardaware wallet the shared secret and decrypt locally) because hadware wallet does not provide access to private key to decrypt data and are to rudimentary to process the decryption by itself.

This PR will permit me to finish EIP1024 decryption for ledger keyring ([https://github.com/MetaMask/eth-ledger-bridge-keyring/](https://github.com/MetaMask/eth-ledger-bridge-keyring/))

Thanks you & ask anything you want !

### DIFF
--- a/src/encryption.test.ts
+++ b/src/encryption.test.ts
@@ -1,8 +1,10 @@
 import {
   decrypt,
-  decryptSafely, decryptWithSharedSecret,
+  decryptSafely,
+  decryptWithSharedSecret,
   encrypt,
-  encryptSafely, EthEncryptedData,
+  encryptSafely,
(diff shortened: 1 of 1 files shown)


LucasALLOIN:
Hello @Gudahtt sorry for pinging you, but looks like you are one of the main maintainer of this repo.
Can we have a news on this ? Any question or remark ?

PR #260, opened 2022-07-27
Replace heavy crypto packages for lighter noble implementations via upgrading `ethereumjs-util` to latest (now called `@ethereumjs/util`)

adonesky1:
Removes `secp256k1` and `keccak` via as dependencies by upgrading from `ethereumjs-util@6.2.1` to its latest (beta) version `@ethereumjs/util@8.0.0-beta.1` which uses noble implementations for crypto. Replaces keccak hashing with `keccak256` from `ethereum-cryptography/keccak`
 
To achieve this I've copied functions from `ethereumjs-abi` (rawEncode and solidityPack) that we use to pack the values before hashing them with the newly added `keccak256` implementation from ethereum-cryptography/keccak, and removed the `ethereumjs-abi` dependency. I've extracted these functions and more or less preserved their implementations ([solidityPack](https://npmfs.com/package/ethereumjs-abi/0.6.8/lib/index.js#L518), [rawEncode](https://npmfs.com/package/ethereumjs-abi/0.6.8/lib/index.js#L345)) minus the references to `ethereumjs/util@6.0.0` methods, which, if left intact [would keep](https://npmfs.com/package/ethereumjs-util/6.0.0/dist/index.js) `secp256k1` and `keccak` in our dependency tree.

Next steps:
- Swap out `bn.js` for `BigInt` (per this comment: https://github.com/MetaMask/eth-sig-util/pull/260#issuecomment-1212443084)


src/sign-typed-data.ts
```diff
@@ -159,7 +161,7 @@ function encodeField(
       'bytes32',
       version === SignTypedDataVersion.V4 && value == null // eslint-disable-line no-eq-null
         ? '0x0000000000000000000000000000000000000000000000000000000000000000'
-        : keccak(encodeData(type, value, types, version)),
+        : toBuffer(keccak256(encodeData(type, value, types, version))),
```
kumavis, line 164: is there a more explicit conversion utility? bigint to buffer or whatever?

src/sign-typed-data.ts
```diff
@@ -168,15 +170,23 @@ function encodeField(
   }
 
   if (type === 'bytes') {
-    return ['bytes32', keccak(value)];
+    if (typeof value === 'number') {
+      value = numberToBuffer(value);
+    } else if (isHexString(value) && parseInt(value, 16) <= 16) {
```
kumavis, line 175: why is this value limited to 16 only for hex strings

src/sign-typed-data.ts
```diff
@@ -459,12 +472,18 @@ function _typedSignatureHash(typedData: TypedDataV1): Buffer {
...
-      soliditySHA3(new Array(typedData.length).fill('string'), schema),
-      soliditySHA3(types, data),
-    ],
+  return toBuffer(
+    keccak256(
+      solidityPack(
+        ['bytes32', 'bytes32'],
+        [
+          keccak256(
+            solidityPack(new Array(typedData.length).fill('string'), schema),
```
kumavis, line 481: weird fill call

src/ethereumjs-abi-utils.ts
```diff
@@ -0,0 +1,390 @@
...
+  setLengthLeft,
+  isHexPrefixed,
+  zeros,
+} from '@ethereumjs/util';
+import { stripHexPrefix } from 'ethjs-util';
+import BN from 'bn.js';
+import { normalize } from './utils';
+
+//
+// Methods borrowed and somewhat adapted from ethereumjs-abi@0.6.8:
```
kumavis, line 18: this is a lot of code - it's not available in a sufficient manner elsewhere?

kumavis (changes requested):
as possible try to remove toBuffer calls where we know the input type

src/sign-typed-data.ts
adonesky1: I think this was an intermediate step that we can remove now...

src/ethereumjs-abi-utils.ts
adonesky1: not that I'm seeing... but can look around more!

src/sign-typed-data.ts
adonesky1: removed here: https://github.com/MetaMask/eth-sig-util/pull/260/commits/ba99a2c2eedbb5f3ec8d6adae387b54eca1272d1

src/sign-typed-data.ts
adonesky1: `keccak256` returns a `Uint8array` so I've swapped this out for `arrToBufArr` from `@ethereumjs/util`

src/ethereumjs-abi-utils.ts
adonesky1: Yeah so for instance I looked into using ethers version of `solidityPack` here https://docs.ethers.io/v5/api/utils/hashing/#utils-solidityKeccak256, which ultimately depends on https://npmfs.com/package/js-sha3/0.8.0/src/sha3.js for its implementation of `Keccak256`

Actually looking at it again, https://www.npmjs.com/package/@ethersproject/solidity might be the way to go to replace `solidityPack`... it has more than we need but would eliminate some of these new utility functions I brought in.

adonesky1:
@kumavis this currently appears to be increasing the build size of the package...

adonesky1:
> @kumavis this currently appears to be increasing the size of the package... 🤦

This is just true of the dist/ folder, not the overall impact on the consumer and its dependency tree, so yeah its gonna be bigger with all those added utility functions

wbt:
Updating ethereumjs-util is also useful for the reasons noted [here](https://github.com/trufflesuite/truffle/issues/5260) around getting consistent types from bn.js after a recent breaking-change update adopted elsewhere.

src/ethereumjs-abi-utils.ts
adonesky1: using the `keccak256` (which uses `pack`) function from `@ethersproject/solidity` is causing a bunch of issues, some of which I have been able to fix, but unsure if this is a worthwhile path, will bring it to you tomorrow at office hours. Would like to get some clarity about measuring the impact here before proceeding.

kumavis: after review, i think its ok to take on the abi code burden

paulmillr:
Why are you keeping `bn.js` if all the internal methods of new `@ethereumjs` libraries switched to native bigints? For keccak256, you can use `ethereum-cryptography/keccak256` since you're already using it.

adonesky1:
> Why are you keeping bn.js if all the internal methods of new @ethereumjs libraries switched to native bigints

I'm using `bn.js` here to replicate the functions from `ethereumjs-abi` (`rawEncode` and `solidityPack`) that we use to pack the values **before** hashing them with the `keccak256` implementation from `ethereum-cryptography/keccak`. I've extracted these functions and more or less preserved their implementations ([solidityPack](https://npmfs.com/package/ethereumjs-abi/0.6.8/lib/index.js#L518), [rawEncode](https://npmfs.com/package/ethereumjs-abi/0.6.8/lib/index.js#L345)) - including use of bn.js - minus the references to old `ethereumjs/util@6.0.0` methods, which [would keep](https://npmfs.com/package/ethereumjs-util/6.0.0/dist/index.js) `secp256k1` and `keccak` in our dependency tree.

paulmillr:
@adonesky1 understood, so it's basically copy-pasting code from other lib. I've noticed about 5 lines that use `BN`. Think it would be really easy to replace it with native `BigInt`. In the current case you're adding 3547 lines of unnecessary code.

adonesky1:
> I've noticed about 5 lines that use BN. Think it would be really easy to replace it with native BigInt. In the current case you're adding 3547 lines of unnecessary code.

Yeah looking into this now. Will try to make it work in a subsequent PR.

kumavis (approved):

PR #263, opened 2022-08-10
Bump Node to v14

bowensanders:
This changes all refs to v12 to point to v14 as v12 is no longer LTS


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [4.0.2]
+### Changed
+- **BREAKING:** Removed support for Node v12 in favor of v14 ([#137](https://github.com/MetaMask/eth-json-rpc-middleware/pull/137))
+
```
rekmarks, line 12: This entry should just go under the `Unreleased` header until we actually cut a release. It's also fine not to modify the changelog at all until a release is cut, because our release automation automatically adds an entry for each commit to the default branch since the previous release.

rekmarks (changes requested):

CHANGELOG.md
bowensanders: I believe I've changed everything to be correct.

.github/workflows/lint-test.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 16.8]
+        node-version: [14.x, 16.8]
```
rekmarks, line 14:
```suggestion
        node-version: [14.x, 16.x, 18.x]
```

rekmarks: The reason we can switch to `16.x` is because the reason for pinning the version no longer applies: https://github.com/MetaMask/eth-sig-util/pull/199

Gudahtt:
It looks like the error message changed between Node.js versions. We could update those tests to just look for the prefix that is common between all versions. That should get tests passing again.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- **BREAKING:** Removed support for Node v12 in favor of v14 ([#137](https://github.com/MetaMask/eth-json-rpc-middleware/pull/137))
```
mcmire, line 9: We can save updating the changelog / bumping the version to the end (the version probably needs to be 5.0.0 anyway).

bowensanders: Just to be clear are you suggesting we leave the above addition to [Unreleased] out at this point?

bowensanders:
> It looks like the error message changed between Node.js versions. We could update those tests to just look for the prefix that is common between all versions. That should get tests passing again.

That's almost finished now. One of the tests is different than the others, and the only one failing, which I am currently working on.

package.json
```diff
@@ -66,7 +66,7 @@
     "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-import": "^2.23.4",
     "eslint-plugin-jest": "^24.3.6",
-    "eslint-plugin-jsdoc": "^36.1.0",
+    "eslint-plugin-jsdoc": "^39.2.0",
```
Gudahtt, line 69: Hmm. This generates peer dependency warnings, because it is outside the range accepted by our ESLint config. I understand that this was required for Node.js v18 compatibility though.

I guess we can ignore the peer dependency warning for now if that's easiest. I can update the ESLint config in a separate PR to resolve it.

CHANGELOG.md
Gudahtt: Removing the version bump is important because our release process relies on that not having been done yet.

This changelog entry is fine to leave here, but it's of limited help. These get auto-generated by our release process.The auto-generated entries aren't formatted as nicely, they need to be manually pruned and cleaned up, but it's enough that it's probably not worth your time doing this going forward.

mcmire: Yeah so just to summarize: leave the changelog entry, but drop the version bump.

bowensanders: done

Gudahtt (approved):
LGTM!

rekmarks (approved):
LGTM!

PR #272, opened 2022-09-22
Fix typo in README

brendankirby:
There is a typo in the README file ("requried") 
The spelling should be updated to "required"


ryanml (approved):
Thanks for this!

PR #273, opened 2022-10-08
Swap out `ethereumjs-abi` for `abi-utils`

Mrtenz:
This removes the inlined `ethereumjs-abi` package in favour of `@metamask/abi-utils`. That also means that we can get rid of the `BN.js` dependency.

It's still a work in progress. The `abi-utils` library is much more strict than `ethereumjs-abi`, so a lot of values are failing to encode right now. There also seems to be a difference in signatures for certain values, which I have yet to figure out.


package.json
```diff
@@ -89,5 +90,8 @@
       "ethereumjs-util>ethereum-cryptography>keccak": true,
       "ethereumjs-util>ethereum-cryptography>secp256k1": true
     }
+  },
+  "resolutions": {
+    "@metamask/abi-utils": "portal:/Users/morten/Development/MetaMask/abi-utils"
```
ritave, line 95: .

Mrtenz: This is temporary while waiting on a new release of `abi-utils`.

PR #276, opened 2022-11-09
Update GitHub actions to latest from module template

Gudahtt:
The GitHub actions have been updated to match the latest module template actions. The main difference is that the `require-additional- reviewer` action has been omitted because it's currently broken for forks.

This adds support for npm publishing, and it improves how the documentation is published. We'll now have a separate subdirectory for the docs for each release, as well as directories for latest stable and staging.


Gudahtt:
I'll have to migrate the docs currently on the `gh-pages` branch to a subdirectory prior to moving this, to avoid any conflicts on that branch.

Mrtenz (changes requested):
Make sure to add a `build:docs` script.

Mrtenz:
I can't see through the GitHub interface if the external scripts are executable, but please make sure that they are too.

Gudahtt:
Thanks @Mrtenz ! Putting this back in draft to investigate how the current `docs:publish` command works, specifically the git revision part.

I have verified that the scripts are executable.

Mrtenz:
> Putting this back in draft to investigate how the current docs:publish command works, specifically the git revision part.

We were using this in other places too, but I can't remember where exactly right now. I'm not sure what it does, but haven't found any issues when replacing it with just `typedoc`.

Gudahtt:
Found it: https://typedoc.org/guides/options/#gitrevision

Looks like it will statically set the revision linked to by the docs, rather than using the latest. That sounds like a good improvement. We don't want to link to the wrong code. I'll leave that in and update the template.

Gudahtt:
I've just confirmed that the default behavior already does the correct thing, so I'll omit both options.

The preserve output directory option was only needed to support the now-deprecated docs publishing action we were using, that's not needed either.

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,148 @@
...
+            exit 1
+          fi
+  check-workflows:
+    name: Check workflows
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Download actionlint
+        id: download-actionlint
+        run: bash <(curl https://raw.githubusercontent.com/rhysd/actionlint/d5f726fb9c9aaff30c8c3787a9b9640f7612838a/scripts/download-actionlint.bash) 1.6.21
```
Mrtenz, line 134: We could use the latest version here. This should be updated in the module template too, will do that soon.
```suggestion
        run: bash <(curl https://raw.githubusercontent.com/rhysd/actionlint/7fdc9630cc360ea1a469eed64ac6d78caeda1234/scripts/download-actionlint.bash) 1.6.22
```

Mrtenz (approved):
LGTM. Just a small suggestion.

.github/workflows/build-test.yml
Gudahtt: Good idea!

Mrtenz (approved):

PR #280, opened 2022-11-10
allow any minor version of bn.js above 4.11.8

adonesky1:
We're going to get rid of `bn.js` with https://github.com/MetaMask/eth-sig-util/pull/273, but until then we should allow any minor version bump above `4.11.8` so that we can consolidate our bundle on on version for all instances of `v4`. Not sure why I didn't think we could do this before.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/eth-sig-util",
-  "version": "5.1.0",
+  "version": "5.0.1",
   "description": "A few useful functions for signing ethereum data",
   "keywords": [
     "ethereum",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/eth-sig-util",
-  "version": "5.0.1",
+  "version": "5.1.0",
```
Gudahtt, line 3: This will be done in the release branch
```suggestion
  "version": "5.0.1",
```

Gudahtt (commented):
LGTM! Can approve after removing the version bump

package.json
adonesky1: Not sure what happened here... I didn't conciously make this change lol

Gudahtt (approved):

PR #301, opened 2023-03-17
Break out domainHash generation

danfinlay:
Allows more easily debugging at least that one component of the 712 signature.


mcmire:
If we could add a test for this new function that would be great. Maybe in a new PR? :)

danfinlay:
The good news is that it has 100% code coverage, since it is already part of every other 712 method. I will see about writing a test, but I’ve been stretched a bit thin.

mcmire:
@danfinlay Oh sure. No problem then!

PR #302, opened 2023-03-17
chore: bump ethereum-cryptography, @ethereumjs/util

legobeat:


mcmire:
@legobeat Are there any consequences for bumping these packages? Is this just to make sure we're up to date, or is there an issue that this fixes?

Mrtenz:
> should be no breaking changes.

`@ethereumjs/util` now uses `@chainsafe/as-sha256` which uses WASM, so it doesn't work in all environments.

package.json
```diff
@@ -39,9 +39,9 @@
     "test:watch": "jest --watch"
   },
   "dependencies": {
-    "@ethereumjs/util": "^8.0.0",
+    "@ethereumjs/util": "8.0.3",
```
mcmire, line 42: Did we want to keep the `^` to allow for future updates?

mcmire:
Good call @Mrtenz.

While we're on the subject, this looks to be the comparison between ethereum-cryptography 1.1.2 and 1.2.0: https://github.com/ethereum/js-ethereum-cryptography/compare/c434a5f24b2c1973cacffd042efe7dcf089aa770...0cf402cc4ec6c6acc505b81f244c0278b366b95a (no tags for these releases, unfortunately, and no changelog either). I'm not spotting anything that could cause any issues, but I'm also not an subject matter expert on this.

FrederikBolding:
@mcmire @legobeat I think we should just hold off for a second on this PR and bump to the versions released here: https://github.com/ethereumjs/ethereumjs-monorepo/pull/2648

I have an open PR here that will most likely be included to fix any incompatibilities with the extension: https://github.com/ethereumjs/ethereumjs-monorepo/pull/2656

Mrtenz:
`@chainsafe/as-sha256` is still added in `yarn.lock`. Is this no longer used?

Mrtenz (approved):

FrederikBolding:
> `@chainsafe/as-sha256` is still added in `yarn.lock`. Is this no longer used?

It is available, but needs to be enabled to be used. By default it will use `noble`

FrederikBolding (approved):

PR #306, opened 2023-04-25
fix: return "0x00" for normalize(0)

legobeat:
The `normalize` function returns `0x${number}` for all numbers except `0`, for which it returns `undefined`. This changes it to return `0x00` for `0`.

### DIFF
--- a/src/utils.test.ts
+++ b/src/utils.test.ts
@@ -111,6 +111,12 @@ describe('normalize', function () {
     expect(result).toBe(`0x${initial.toLowerCase()}`);
   });
 
+  it('should normalize 0 to a byte-pair hex string', function () {
+    const initial = 0;
+    const result = normalize(initial);
+    expect(result).toBe('0x00');
+  });
+
   it('should normalize an integer to a byte-pair hex string', function () {
     const initial = 1;
     const result = normalize(initial);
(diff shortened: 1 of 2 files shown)


kumavis (dismissed):
good

this is a breaking change

Gudahtt (approved):
LGTM!

mikesposito:
I didn't notice this PR before opening this one: https://github.com/MetaMask/eth-sig-util/pull/315
@legobeat I am thinking if maybe #315 could solve the problem you are trying to fix?

mcmire:
@mikesposito @legobeat This seems complementary to #315. I don't see a reason why we couldn't have both. Do you?

mcmire (approved):
Makes sense.

mikesposito:
@mcmire We can definitely have both, I was just pointing out that after #315 has been merged, the early function termination does not affect `0` input anymore, so from this PR we could also just add the test

Gudahtt:
I don't think the functional change made in this PR is relevant anymore. The change in #315 is functionally equivalent in the case where zero is passed in, though broader in that it also affects the case where an empty string is passed in.

Agreed that the test would be useful to include though!

Gudahtt:
I have created a separate PR to add just the test: https://github.com/MetaMask/eth-sig-util/pull/317

Gudahtt:
Superseded by #315 and #317

PR #307, opened 2023-04-25
update eslint; types/lint fixes

legobeat:
- update `eslint`@`7.30.0`->`8.27.0`
  - bump eslint plugins and configs
    - contains #290 
    - contains #291 
    - contains #292 
    - contains #293 
- add missing type declarations
  - some gaps became apparent but saving changes to code beyond the minimum for later changes. 
- add devDependency `@types/bn.js`

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,12 @@
   },
   "main": "./dist/index.js",
   "files": [
-    "dist"
+    "dist",
+    "!__snapshots__",
+    "!*.test.js",
+    "!*.test.js.map",
+    "!*.test.ts",
+    "!*.test.d.ts"
   ],
   "scripts": {
     "build": "tsc --project .",


src/encryption.ts
```diff
@@ -166,7 +166,7 @@ export function decrypt({
   switch (encryptedData.version) {
     case 'x25519-xsalsa20-poly1305': {
       // string to buffer to UInt8Array
-      const recieverPrivateKeyUint8Array = nacl_decodeHex(privateKey);
+      const recieverPrivateKeyUint8Array = naclDecodeHex(privateKey);
       const recieverEncryptionPrivateKey = nacl.box.keyPair.fromSecretKey(
```
kumavis, line 170: typo

src/ethereumjs-abi-utils.test.ts
```diff
@@ -87,23 +88,39 @@ describe('encoding', function () {
...
+    expect(() => rawEncode('fail' as any, ['bytes33'])).toThrow(
       'types.forEach is not a function',
     );
   });
 });
 
+describe('encoding invalid type', function () {
+  it('should fail', function () {
+    expect(() => rawEncode(['fail'], ['bytes33'])).toThrow(
+      'Unsupported or invalid type: fail',
```
kumavis, line 100: wrap in quotes if possible

src/ethereumjs-abi-utils.ts
```diff
@@ -58,7 +60,7 @@ function isArray(type) {
  * @param type - A string that may be an array to parse.
  * @returns A parsed value from the array.
  */
-function parseTypeArray(type) {
+function parseTypeArray(type: string): 'dynamic' | number | null {
```
kumavis, line 63: consider wrapping type sets { 'dynamic' | number | null } as a new type

kumavis (dismissed):
nice type improvements

src/ethereumjs-abi-utils.ts
```diff
@@ -82,18 +88,20 @@ function parseTypeN(type) {
  * @param arg - Number to parse.
  * @returns Parsed value.
  */
-export function parseNumber(arg) {
+export function parseNumber(arg: string | number | BN): BN {
   const type = typeof arg;
   if (type === 'string') {
```
Mrtenz, line 93: If you do `typeof arg` here, you can remove the type cast further on.
```suggestion
  if (typeof arg === 'string') {
```

src/ethereumjs-abi-utils.ts
```diff
@@ -82,18 +88,20 @@ function parseTypeN(type) {
...
-    if (isHexPrefixed(arg)) {
+    if (isHexPrefixed(arg as string)) {
```
Mrtenz, line 94:
```suggestion
    if (isHexPrefixed(arg)) {
```

src/ethereumjs-abi-utils.ts
```diff
@@ -82,18 +88,20 @@ function parseTypeN(type) {
...
       return new BN(stripHexPrefix(arg), 16);
     }
     return new BN(arg, 10);
   } else if (type === 'number') {
     return new BN(arg);
-  } else if (arg.toArray) {
-    // assume this is a BN for the moment, replace with BN.isBN soon
-    return arg;
+  } else if (
+    (arg && Object.prototype.hasOwnProperty.call(arg, 'toArray')) ||
```
Mrtenz, line 101: Do we still need this now that we use `BN.isBN`?

tsconfig.json
```diff
@@ -9,9 +9,15 @@
...
     "typeRoots": ["./node_modules/@types"]
   },
-  "include": ["./src/**/*.ts"],
-  "exclude": ["**/*.test.ts"]
+  "jsonCompilerOptions": {
+    "strictNullChecks": true
+  },
+  "parserOptions": {
+    "project": ["./tsconfig.json"]
+  },
```
Mrtenz, line 21: What's this for?

Mrtenz (changes requested):

src/ethereumjs-abi-utils.ts
Mrtenz: Fair. This whole file will be removed in #273 regardless. 😅

Mrtenz (approved):

Mrtenz (approved):

.eslintrc.js
```diff
@@ -12,6 +12,7 @@ module.exports = {
     },
   ],
   rules: {
+    '@typescript-eslint/prefer-nullish-coalescing': 'off',
```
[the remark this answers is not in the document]
Gudahtt: It looks like this is being set in the wrong place. This `rules` property is for the _base_ rules, applied to the entire repository. The TypeScript rule you're trying to disable here is set in an override block, so that it only applies to TypeScript files. The overrides will override anything set here.

You can disable this by moving this into a `rules` block _inside_ the TypeScript override block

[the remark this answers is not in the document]
Gudahtt: Though... I am curious, why disable this rule?

package.json
```diff
@@ -49,22 +49,23 @@
...
     "@lavamoat/allow-scripts": "^2.0.3",
     "@metamask/auto-changelog": "^3.1.0",
-    "@metamask/eslint-config": "^9.0.0",
-    "@metamask/eslint-config-jest": "^9.0.0",
-    "@metamask/eslint-config-nodejs": "^9.0.0",
-    "@metamask/eslint-config-typescript": "^9.0.1",
+    "@metamask/eslint-config": "^11.1.0",
+    "@metamask/eslint-config-jest": "^11.1.0",
+    "@metamask/eslint-config-nodejs": "^11.1.0",
+    "@metamask/eslint-config-typescript": "^11.1.0",
```
Gudahtt, line 55: This config version depends upon TypeScript ~4.8, but currently 4.7.4 is used in the lockfile. Perhaps we should update the TypeScript version here too (and switch to a `~` range to prevent accidental breaking changes going forward).

tsconfig.json
```diff
@@ -9,9 +9,15 @@
     "outDir": "dist",
     "rootDir": "src",
     "sourceMap": true,
+    "strictNullChecks": true,
     "target": "ES2017",
     "typeRoots": ["./node_modules/@types"]
   },
-  "include": ["./src/**/*.ts"],
-  "exclude": ["**/*.test.ts"]
```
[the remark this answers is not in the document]
Gudahtt: Try running `npm publish --dry-run` - it shows that the npmignore file is not working (presumably because `files` is set and that takes precedence).

You could introduce the pattern from the module template where we have a separate `tsconfig.build.json` for the production build which excludes tests, but leave tests included in the top-level one.

Or if you'd prefer, we could update `files` to use a glob that excludes tests

tsconfig.json
```diff
@@ -9,9 +9,12 @@
...
+  "jsonCompilerOptions": {
```
Gudahtt, line 16: I am unfamiliar with this option - what does this do?

.eslintrc.js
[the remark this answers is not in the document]
Gudahtt: Cool, makes sense

src/encryption.test.ts
```diff
@@ -154,7 +154,7 @@ describe('encryption', function () {
       it('should throw if passed null public key', function () {
         expect(() =>
           encrypt({
-            publicKey: null,
+            publicKey: null as any,
```
Gudahtt, line 157: Nit: We could use `// @ts-expect-error` in each of these places rather than casting to `any`. It's a bit safer in that we get notified if the type error goes away.

It also prevents the `any` type "leaking"  onto other lines where the type checking would still be valuable (not applicable here because `null` is a literal, but it might apply if this was done to a variable).

src/encryption.ts
```diff
@@ -183,21 +183,21 @@ export function decrypt({
...
-        output = naclUtil.encodeUTF8(decryptedMessage);
+        if (!decryptedMessage) {
+          throw new Error();
+        }
+        return naclUtil.encodeUTF8(decryptedMessage);
       } catch (err) {
-        throw new Error('Decryption failed.');
-      }
-
-      if (output) {
```
Gudahtt, line 197: This had the effect of throwing the "Decryption failed." error if the resulting output was an empty string. Perhaps that's a behavior we should preserve in this PR at least (I'm not entirely sure why we'd want to throw in that circumstance but :shrug: )

src/utils.ts
```diff
@@ -117,7 +117,7 @@ export function normalize(input: number | string): string {
 
   if (typeof input !== 'string') {
     let msg = 'eth-sig-util.normalize() requires hex string or integer input.';
-    msg += ` received ${typeof input}: ${input}`;
+    msg += ` received ${typeof input}: ${input as any as string}`;
```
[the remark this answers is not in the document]
Gudahtt: My bad, my ESLint process has crashed and I hadn't noticed :facepalm: It does indeed throw an error as you'd expect.

tsconfig.json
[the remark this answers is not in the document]
Gudahtt: Looks like this is still a problem; tests are being included in the package output

Gudahtt:
Mostly looks good! I can approve once [the issue with tests being published](https://github.com/MetaMask/eth-sig-util/pull/307#discussion_r1180363151) has been fixed.

Gudahtt (approved):

PR #315, opened 2023-06-22
Fix `normalize` for empty strings

mikesposito:
Currently the `normalize` function uses a logical NOT (!) to check if the input is nullish, but this causes to return `undefined` when an empty string is passed as input.

As a fix, this PR uses the `isNullish` function instead of the logical NOT.

## Changes

* **FIXED**: `normalize` function for empty strings

## References

* See [#1441](https://github.com/MetaMask/core/pull/1441#issuecomment-1602515812)
* Related to [#1315](https://github.com/MetaMask/core/issues/1315)

### DIFF
--- a/src/utils.test.ts
+++ b/src/utils.test.ts
@@ -124,6 +124,12 @@ describe('normalize', function () {
     expect(result).toBe('0x');
   });
 
+  it('should normalize an empty string to 0x', function () {
+    const initial = '';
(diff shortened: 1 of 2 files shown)


mcmire (commented):
Can we add a test for this?

mcmire (approved):
Looks good!

PR #317, opened 2023-06-29
Add test for normalizing zero

Gudahtt:
A test has been added to capture the behavior of the `normalize` function when a zero is passed in. This behavior changed recently in PR #315.


Gudahtt:
This test was ported from #306

mikesposito (approved):
LGTM!

Issue #289, opened 2022-12-14
Throw away tweetnacl and tweetnacl-util


paulmillr:
There's faster noble/ed25519  which is also audited; also it's used already through MM dependency tree

PR #358, opened 2023-11-22
deps: replace tweetnacl-util with @scure/base

legobeat:
- `tweetnacl-util` is no longer actively maintained
- `@scure/base` is already pulled in transitively through `ethereum-cryptography`
- Remove redundant base64 encode/decode step when decoding hex input

### Related
- #289

### DIFF
--- a/src/encryption.ts
+++ b/src/encryption.ts
@@ -1,5 +1,5 @@
+import { base64, utf8 } from '@scure/base';
 import * as nacl from 'tweetnacl';
-import * as naclUtil from 'tweetnacl-util';
 
 import { isNullish } from './utils';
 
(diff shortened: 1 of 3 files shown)


src/encryption.ts
```diff
@@ -165,18 +165,15 @@ export function decrypt({
 
   switch (encryptedData.version) {
     case 'x25519-xsalsa20-poly1305': {
-      // string to buffer to UInt8Array
-      const receiverPrivateKeyUint8Array = naclDecodeHex(privateKey);
+      const receiverPrivateKeyUint8Array = Buffer.from(privateKey, 'hex');
```
Mrtenz, line 168: Can we use `@metamask/utils` instead of Node.js globals?

legobeat: Can we address that in a separate PR? This function is already using the native `Buffer` ([here](https://github.com/MetaMask/eth-sig-util/pull/358/files#diff-e15805cccaeceb899db45f9ad7fa014fe8373c838e340e85d29f967afb4d0eb8L258)), so we're not introducing it here.

Mrtenz (approved):

PR #367, opened 2024-05-01
test(encryption): add compatibility-test to v7.0.1

legobeat:
Encryption in this library is historically using `tweetnacl` and `tweetnacl-util` as encryption libraries. `tweetnacl-util` was replaced with `@scure/base` in #358 and `tweetnacl` is also due for removal.

This adds a snapshot of `encryption.ts` as `test-legacy-encryption.ts`, and adds compatibility tests in both directions to ensure that new versions of the code can decrypt messages produced by the old version and vice-versa.


src/encryption.test.ts
```diff
@@ -5,349 +5,390 @@ import {
...
-
-      it('should throw if passed undefined encrypted data', function () {
-        expect(() =>
-          decryptSafely({
-            encryptedData: undefined as any,
-            privateKey: bob.ethereumPrivateKey,
-          }),
-        ).toThrow('Missing encryptedData parameter');
-      });
+run({
```
mcmire, line 372: I'm curious whether it would be obvious that are we are running these tests with different inputs, or even why we are doing so. What are your thoughts on using `describe` to organize/describe the different invocations? I'm thinking an arrangement that looks something like this:

``` typescript
describe('encryption', () => {
  describe('using modern encryption and decryption APIs', () => {
    run({
      decrypt,
      decryptSafely,
      encrypt,
      encryptSafely,
      getEncryptionPublicKey,
    });
  });

  describe('using legacy encryption API', () => {
    run({
      decrypt,
      decryptSafely,
      encrypt: legacyEncrypt,
      encryptSafely: legacyEncryptSafely,
      getEncryptionPublicKey: legacyGetEncryptionPublicKey,
    });
  });

  describe('using legacy decryption API', () => {
    run({
      decrypt: legacyDecrypt,
      decryptSafely: legacyDecryptSafely,
      encrypt,
      encryptSafely,
      getEncryptionPublicKey,
    });
  });
});
```

Note that this would mean removing the `describe('encryption', ...)` from `run`.

PR #390, opened 2024-10-03
drop support for Node.js versions 16, 21

legobeat:
- Remove support for EoL Node.js versions 16, 21
- Align version of `@types/node`
- ci: Only run lint on default Node version
  - ci: Bump @actions/setup-node to v4
  - ci: Simplify caching of yarn node modules
- ci: Run on Node.js v22

#### Blocking
- https://github.com/MetaMask/utils/pull/212

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -14,11 +14,14 @@ jobs:
       YARN_VERSION: ${{ steps.yarn-version.outputs.YARN_VERSION }}
     strategy:
       matrix:
-        node-version: [16.x, 18.x, 20.x]
+        node-version:
+          - 18.x
+          - 20.x
+          - 22.x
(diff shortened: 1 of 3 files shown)


ccharly:
We have been moving all accounts-related packages to a [new monorepo here](https://github.com/MetaMask/accounts).

While doing the migration, we had to unify the build workflows, and we did drop node16 support starting with the very first release of the monorepo:
- https://github.com/MetaMask/accounts/blob/v1.0.0/.github/workflows/build-lint-test.yml
- https://github.com/MetaMask/accounts/releases/tag/v1.0.0

So I believe you could just use those new versions of each packages, and that might resolve those blockers.

Let me know if you have any question!

cryptodev-2s (approved):

PR #403, opened 2024-12-09
Add `Permit` type schema

OGPoyraz:
### Description

This PR aims to extend `Permit` type in `TYPED_MESSAGE_SCHEMA`.

If the `value` field is present in `Permit` primary type schemas it must be string, otherwise it will not check the type.


cryptotavares (changes requested):
We cannot make this assumptions. This would make a perfectly valid eip712 signature, invalid.

OGPoyraz:
@cryptotavares this is not making `value` to be `required`, can you give an example where this will break?

PR #407, opened 2024-12-19
Adds `signAuthorization` method for EIP-7702

jeffsmale90:
EIP-7702 defines a new struct `Authorization` which represents authority to set a pointer to a contract address at an EOA - effectively making the EOA perform as a smart contract.

This change adds methods to hash and sign authorizations, and to recover signer address.

Although the EIP defines the signed authorization as `[chain_id, address, nonce, y_parity, r, s]`, the utility method returns the signature as a standard `r|s|v` signature for consistency with other signing methods.

See https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7702.md for details

### DIFF
--- a/src/sign-authorization.test.ts
+++ b/src/sign-authorization.test.ts
@@ -1,4 +1,5 @@
 import { bufferToHex, privateToAddress } from '@ethereumjs/util';
+
 import {
   signAuthorization,
   recoverAuthorization,
@@ -118,7 +119,7 @@ describe('signAuthorization', () => {
     it('should produce the correct hash', () => {
       const hash = hashAuthorization(testAuthorization);
 
-      expect(hash).toEqual(expectedAuthorizationHash);
+      expect(hash).toStrictEqual(expectedAuthorizationHash);
     });
 
     it('should throw if authorization is null', () => {
@@ -213,16 +214,12 @@ describe('signAuthorization', () => {
 
   describe('sign-and-recover', () => {
     const testCases = {
-      'zero chainId': [0, '0x1234567890123456789012345678901234567890', 1],
-      'high chainId': [98765, '0x1234567890123456789012345678901234567890', 1],
-      'zero nonce': [8545, '0x1234567890123456789012345678901234567890', 0],
-      'high nonce': [8545, '0x1234567890123456789012345678901234567890', 98765],
-      'zero contractAddress': [
-        1,
-        '0x0000000000000000000000000000000000000000',
-        1,
-      ],
-      'all zero values': [0, '0x0000000000000000000000000000000000000000', 0],
+      zeroChainId: [0, '0x1234567890123456789012345678901234567890', 1],
+      highChainId: [98765, '0x1234567890123456789012345678901234567890', 1],
+      zeroNonce: [8545, '0x1234567890123456789012345678901234567890', 0],
+      highNonce: [8545, '0x1234567890123456789012345678901234567890', 98765],
+      zeroContractAddress: [1, '0x0000000000000000000000000000000000000000', 1],
+      allZeroValues: [0, '0x0000000000000000000000000000000000000000', 0],
     } as { [key: string]: Authorization };
 
     for (const [label, authorization] of Object.entries(testCases)) {
--- a/src/sign-authorization.test.ts
+++ b/src/sign-authorization.test.ts
@@ -1,0 +1,244 @@
+import { bufferToHex, privateToAddress } from '@ethereumjs/util';
+import {
+  signAuthorization,
+  recoverAuthorization,
+  Authorization,
+  hashAuthorization,
+} from './sign-authorization';
+
+const testPrivateKey = Buffer.from(
+  '4af1bceebf7f3634ec3cff8a2c38e51178d5d4ce585c52d6043e5e2cc3418bb0',
+  'hex',
+);
+
+const testAddress = bufferToHex(privateToAddress(testPrivateKey));
+
+const testAuthorization: Authorization = [
+  8545,
+  '0x1234567890123456789012345678901234567890',
+  1,
+];
+
+const expectedAuthorizationHash = Buffer.from(
+  'b847dee5b33802280f3279d57574e1eb6bf5d628d7f63049e3cb20bad211056c',
+  'hex',
+);
+
+const expectedSignature =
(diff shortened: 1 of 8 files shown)


src/sign-authorization.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  if (isNullish(nonce)) {
+    throw new Error('Missing nonce parameter');
+  }
+}
+
+/**
+ * Sign an authorization message with the provided private key.
+ * The signature format follows EIP-155 for chain-specific signatures.
```
jeffsmale90, line 49:
```suggestion
```

src/sign-authorization.ts
```diff
@@ -0,0 +1,127 @@
...
+import { concatSig, isNullish, recoverPublicKey } from './utils';
+
+/**
+ * The authorization struct as defined in EIP-7702.
+ *
+ * @property chainId - The chain ID or 0 for any chain.
+ * @property contractAddress - The address of the contract being authorized.
+ * @property nonce - The nonce of the signing account (at the time of submission).
+ */
+export type Authorization = [
```
matthewwalsh0, line 15: Minor, could `Authorization` be or become too generic a term in the context of signing?

Should we call this something like `EIP7702Authorization`?
Maybe `7702` as the filename?

src/sign-authorization.ts
```diff
@@ -0,0 +1,127 @@
...
+  nonce: number,
+];
+
+/**
+ * Validates an authorization object to ensure all required parameters are present.
+ *
+ * @param authorization - The authorization object to validate.
+ * @throws {Error} If the authorization object or any of its required parameters are missing.
+ */
+function validateAuthorization(authorization: Authorization) {
```
matthewwalsh0, line 27: Minor, non-exported functions at the bottom of the file to highlight what is used externally?

src/sign-authorization.ts
```diff
@@ -0,0 +1,127 @@
...
+  if (isNullish(authorization)) {
+    throw new Error('Missing authorization parameter');
+  }
+
+  const [chainId, contractAddress, nonce] = authorization;
+
+  if (isNullish(chainId)) {
```
matthewwalsh0, line 34: Do we definitely want to support / encourage cross-chain authorizations via the `0` chain ID?

Could we also validate it's not zero at this package level?

src/sign-authorization.ts
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * Sign an authorization message with the provided private key.
+ *
+ * @param options - The signing options.
+ * @param options.privateKey - The private key to sign with.
+ * @param options.authorization - The authorization data to sign.
+ * @returns The '0x'-prefixed hex encoded signature.
+ */
+export function signAuthorization({
```
matthewwalsh0, line 55: Minor, maybe `sign7702Authorization` and the same format for the other functions?

Just to add more clarity if we get other types of authorization in future?

src/sign-authorization.test.ts
```diff
@@ -0,0 +1,241 @@
...
+  '0x1234567890123456789012345678901234567890',
+  1,
+];
+
+const expectedAuthorizationHash = Buffer.from(
+  'b847dee5b33802280f3279d57574e1eb6bf5d628d7f63049e3cb20bad211056c',
+  'hex',
+);
+
+const expectedSignature =
```
matthewwalsh0, line 28: Minor, capitals and underscores for constants so `EXPECTED_SIGNATURE` etc?

src/sign-authorization.test.ts
```diff
@@ -0,0 +1,241 @@
...
+  '0xebea1ac12f17a56a514dfecbcbc8bbee7b089fa3fcee31680d1e2c1588f623df7973cab74e12536678995377da38c96c65c52897750b73462c6760ef2737dba41b';
+
+describe('signAuthorization', () => {
+  describe('signAuthorization()', () => {
+    it('should produce the correct signature', () => {
```
matthewwalsh0, line 33: Minor, `should` is redundant in test names?  `produces the correct signature` for example.

src/sign-authorization.test.ts
```diff
@@ -0,0 +1,241 @@
...
+    const testCases = {
+      zeroChainId: [0, '0x1234567890123456789012345678901234567890', 1],
+      highChainId: [98765, '0x1234567890123456789012345678901234567890', 1],
+      zeroNonce: [8545, '0x1234567890123456789012345678901234567890', 0],
+      highNonce: [8545, '0x1234567890123456789012345678901234567890', 98765],
+      zeroContractAddress: [1, '0x0000000000000000000000000000000000000000', 1],
+      allZeroValues: [0, '0x0000000000000000000000000000000000000000', 0],
+    } as { [key: string]: Authorization };
+
+    for (const [label, authorization] of Object.entries(testCases)) {
```
matthewwalsh0, line 225: Minor, could also use a `it.each`.

matthewwalsh0 (changes requested):

src/sign-authorization.ts
jeffsmale90: I am not sure whether or not we want to support chainId 0 - there's an argument for it, even though it's problematic (nonce likely won't match across all chains).

My preference is to support it at the utility level, and make the decision at the API level.

src/sign-authorization.ts
jeffsmale90: Yeah, I think that makes sense - it's already a bit of an overloaded term.

src/sign-authorization.test.ts
jeffsmale90: You are correct - it could be removed, but `should` is consistent across all of the existing tests 🤷

src/sign-authorization.ts
jeffsmale90: I named the file `sign-eip7702-authorization.ts` for a combination of clarity and consistency with the other signing files.

jeffsmale90:
@matthewwalsh0 thanks for the review! I made all of the changes you recommended, except for validating chainId !== 0, as I think this should be an API concern, rather than a utility concern.

src/sign-authorization.test.ts
matthewwalsh0: It's a legacy pattern in lots of repositories that we're slowing moving aways from, but not urgent now.

matthewwalsh0 (approved):

PR #410, opened 2025-02-05
Add additional validation to EIP7702 methods

jeffsmale90:
Adds validation of the Authorization object, as per EIP-7702 specification https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7702.md

Also renames some tests to align with renamed function names.

### DIFF
--- a/src/sign-eip7702-authorization.test.ts
+++ b/src/sign-eip7702-authorization.test.ts
@@ -216,6 +216,19 @@ describe('signAuthorization', () => {
       ).toThrow('Invalid contractAddress: must be a 20 byte hex string');
     });
 
+    it('should throw if contractAddress is too long', () => {
+      expect(() =>
+        signEIP7702Authorization({
+          privateKey: TEST_PRIVATE_KEY,
+          authorization: [
+            TEST_AUTHORIZATION[0],
+            `${TEST_AUTHORIZATION[1]}00`,
+            TEST_AUTHORIZATION[2],
+          ],
+        }),
+      ).toThrow('Invalid contractAddress: must be a 20 byte hex string');
+    });
+
     it('should throw if contractAddress is not valid hex', () => {
       expect(() =>
         signEIP7702Authorization({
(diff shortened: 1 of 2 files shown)


src/sign-eip7702-authorization.ts
```diff
@@ -126,4 +130,30 @@ function validateEIP7702Authorization(authorization: EIP7702Authorization) {
...
+    typeof nonce !== 'number' ||
+    nonce >= 2 ** NONCE_MAX_BITLENGTH ||
+    nonce < 0
+  ) {
+    throw new Error(
+      `Invalid nonce: must be a non-negative number less than 2^${NONCE_MAX_BITLENGTH}`,
+    );
+  }
+
+  if (!isValidHexAddress(contractAddress as Hex)) {
```
ccharly, line 154: For the record: This function does verify the address length too.

See: https://github.com/MetaMask/utils/blob/v11.3.0/src/hex.ts#L15-L18 (`{40}` in hex-format, each bytes is 2-digits/chars long, which makes it a length of 20 bytes).

src/sign-eip7702-authorization.test.ts
```diff
@@ -88,17 +88,45 @@ describe('signAuthorization', () => {
...
+        'Invalid chainId: must be a non-negative number less than 2^256',
+      );
+    });
+
+    it('should throw if chainId is too large', () => {
+      expect(() =>
+        signEIP7702Authorization({
+          privateKey: TEST_PRIVATE_KEY,
+          authorization: [
+            2 ** 257,
```
ccharly, line 111: Very nit: We could have used `2 ** 256` since that's the upper bound for that range too (which would have "strictly" matched this `assert auth.chain_id < 2**256` from the EIP)

ccharly (dismissed):
Code LGTM! I've left 2 small comments for the tests, but the actual logic is good, so up to you if you want to change this or not. 😄 thanks!

ccharly (approved):

PR #421, opened 2026-08-25
ci: migrate to OIDC trusted publishing and align workflows with module template

OGPoyraz:
## What is the current state of things and why does it need to change?

`@metamask/eth-sig-util`'s CI/publish setup had drifted from the [metamask-module-template](https://github.com/MetaMask/metamask-module-template):

- npm publishing used `MetaMask/action-npm-publish@v1` (token-based) with no `id-token: write`, so it could not use OIDC. Core Platform is enforcing [trusted publishing via OIDC](https://docs.npmjs.com/trusted-publishers) and [staged publishing](https://docs.npmjs.com/staged-publishing/), and non-staged publishes will be rejected.
- The workflows used manual `checkout` / `corepack enable` / `setup-node` / `yarn install` boilerplate instead of the shared `action-checkout-and-setup` action and the reusable `main.yml` orchestrator pattern.
- `packageManager: yarn@3.2.2` was below what `action-npm-publish@v6` requires.

> **Note:** The required Node.js bump (Node >= 22.14.0) landed separately in #422 and is already on `main`.

## What is the solution your changes offer and how does it work?

Aligns the GitHub Actions setup with the module template and adopts OIDC trusted publishing:

- **Adopt OIDC trusted publishing**: upgrade `action-npm-publish` `@v1` → `@v6` (dry-run + publish), add `id-token: write`, and make `NPM_TOKEN` optional (only needed for the first publish, then deleted from the `npm-publish` environment).
- **Switch Yarn to Corepack**: bump `packageManager` to `yarn@4.16.0` (integrity hash), remove `yarnPath` and the committed `.yarn/releases` binary.
- **Restructure workflows to match the template**: add a `main.yml` orchestrator (runs `check-workflows` + `build-lint-test` on push/PR, gates the release via `is-release` → `publish-release`); convert `build-lint-test.yml` and `publish-release.yml` to reusable `workflow_call` workflows using `MetaMask/action-checkout-and-setup@v3` and artifact-based build/publish; update `create-release-pr.yml` (`action-create-release-pr@v5`), `publish-docs.yml`, and `publish-rc-docs.yml`.
- **Bump `@metamask/auto-changelog`** `^3.1.0` → `^6.2.1` so `auto-changelog validate` works on the Node 22/24/26 CI matrix.
- Regenerate `yarn.lock` in the Yarn 4 format. No dependency versions changed beyond `auto-changelog`.

### Intentional adaptations (vs a literal template copy)

- Keep the existing standalone `security-code-scanner.yml` (not folded into `main.yml`).
- Keep `GITHUB_TOKEN` for docs deploys (no `PUBLISH_DOCS_TOKEN` dependency).
- Omit the Slack integration (the repo has no `SLACK_WEBHOOK_URL` secret).
- Use `auto-changelog` for changelog validation (no `lint:changelog` script).
- Keep a plain-string `release-type` input (no `type: choice`), which avoids needing an `actionlint.yml` config file.

Once merged, trusted publishing must be configured for `@metamask/eth-sig-util` on the NPM side (via [@metamask-npm-publishers](https://consensys.slack.com/?subteam=S042S7RE4AE)) before the next release.

## References

- Prerequisite (merged): #422 (drop Node.js 18/20)
- Reference diff: [metamask-module-template@193ed00](https://github.com/MetaMask/metamask-module-template/commit/193ed00ff5b34a20197a6c4a2852e7d9973d413b)
- [NPM trusted publishing docs](https://docs.npmjs.com/trusted-publishers), [staged publishing docs](https://docs.npmjs.com/staged-publishing/)

---

> [!NOTE]
> **High Risk**
> Changes npm publish flow (OIDC, artifacts, job ordering) and release gating on main, so a misconfiguration could block or mishandle releases.
> 
> **Overview**
> Replaces the monolithic **Build, Lint, and Test** workflow with a **`main.yml` orchestrator** on push/PR that runs **actionlint** (new matcher + config, including an ignore for intentional empty `release-type`), a **reusable security scan** (replacing the standalone scanner workflow), and a **`workflow_call` build/lint/test** pipeline using **`MetaMask/action-checkout-and-setup@v3`**. Build and lint now run on **Node 26.x** only (tests still matrix 22/24/26), and a new **compatibility-test** job reinstalls deps without the lockfile before restoring it.
> 
> **Release publishing** is no longer triggered directly from `publish-release.yml` on `main` pushes; **`main.yml`** gates **`is-release`** → **`publish-release.yml`** (now callable). That workflow **builds once, uploads artifacts**, runs **npm dry-run + publish via `action-npm-publish@v6`** with **`id-token: write`**, optional **`NPM_TOKEN`**, then docs deploys and a separate **`action-publish-release@v3`** GitHub release step. **`create-release-pr`** moves to the shared checkout action and **`action-create-release-pr@v5`**, with **`release-type` as a choice** (including empty). Docs publishing adds **`github-pages` environment**, shared checkout/setup, and bumps **peaceiris/actions-gh-pages**. Adds **`.github/zizmor.yml`** for workflow lint policy tweaks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8b88ff0d45387fdc7e0b5b36bd36beca0b235c05. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/workflows/publish-release.yml
```diff
@@ -49,39 +51,54 @@ jobs:
...
           path: ./dist
           key: ${{ github.sha }}
-        # Set `ignore-scripts` to skip `prepublishOnly` because the release was built already in the previous job
-      - run: npm config set ignore-scripts true
       - name: Dry Run Publish
-        # omit npm-token token to perform dry run publish
-        uses: MetaMask/action-npm-publish@v1
+        uses: MetaMask/action-npm-publish@v6
+        with:
+          dry-run: true
```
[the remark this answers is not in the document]
OGPoyraz: Good catch — confirmed and fixed in 0e6c60c.

Reproduced locally: `SKIP_PREPACK=true yarn pack` fails with `YN0001: UsageError: Couldn't find the node_modules state file` when only `./dist` is restored, because `action-npm-publish@v6` runs `yarn pack`/`yarn npm publish` against the `node-modules` linker, which needs `node_modules/.yarn-state.yml`. `SKIP_PREPACK` only skips the rebuild; it does not remove Yarn's need for the install state.

Fix: cache and restore `./node_modules` alongside `./dist` (keyed on the release SHA) in the build, dry-run, and publish jobs. Verified that `yarn pack` then succeeds.

package.json
```diff
@@ -79,9 +79,9 @@
     "typedoc": "^0.24.6",
     "typescript": "~4.8.4"
   },
-  "packageManager": "yarn@3.2.2",
+  "packageManager": "yarn@4.16.0+sha256.ba05224324578801b9cc98170d64aa50b9a36733b440fb0942306da3fbbdc7d1",
   "engines": {
-    "node": "^18.18 || ^20.14 || >=22"
+    "node": "^20 || ^22 || >=24"
```
Mrtenz: This is a breaking change. You should create a separate PR for this, and bump to `^22 || ^24 || >=26` matching the module template.

.nvmrc
```diff
@@ -1 +1 @@
-v18
```
Mrtenz: We use `lts/*` in the module template now. You can do that as part of the separate Node.js bump PR though.

.github/workflows/build-test.yml
```diff
@@ -12,11 +12,12 @@ jobs:
     strategy:
       matrix:
         node-version:
-          - 18.x
           - 20.x
           - 22.x
     steps:
       - uses: actions/checkout@v3
+      - name: Enable Corepack
+        run: corepack enable
```
Mrtenz: Might be a good time to align these more with the module template too. We use `action-checkout-and-setup` which removes the need for manual checkout, corepack, setup Node.js, install dependencies.

package.json
OGPoyraz: Done — split out into #422 (`chore!: drop support for Node.js 18 and 20`), which bumps `engines.node` to `^22 || ^24 || >=26` to match the module template and adds a `**BREAKING**` changelog entry. That PR is merged, and this one has been rebased on top of `main`, so the engines change no longer appears here.

.nvmrc
OGPoyraz: Done in #422 — `.nvmrc` is now `lts/*`. Rebased here, so `.nvmrc` is no longer part of this PR.

.github/workflows/build-test.yml
OGPoyraz: Good call — `action-checkout-and-setup` would remove the manual checkout / `corepack enable` / setup-node / install boilerplate across these workflows (and eliminate the `corepack enable` steps I added here). To keep this PR focused on the OIDC migration and avoid a broad workflow refactor, I have kept the minimal changes for now. Happy to align the workflows with the module template (`action-checkout-and-setup`) — would you prefer I fold that into this PR, or track it as a separate follow-up?

Mrtenz: Fine to make this one big "align with module template" PR. You should mostly be able to just 1:1 copy the workflows (and other required files) from the template.

.github/workflows/build-lint-test.yml
```diff
@@ -0,0 +1,98 @@
...
+          persist-credentials: false
+          node-version: ${{ matrix.node-version }}
+      - run: yarn test
+      - name: Require clean working directory
+        shell: bash
+        run: |
+          if ! git diff --exit-code; then
+            echo "Working tree dirty at end of job"
+            exit 1
+          fi
```
mcmire, line 98: The module template adds a `compatibility-test` job to `build-lint-test.yml` which ensures that minor and patch bumps to dependencies don't accidentally cause breaking changes (ensuring that if the package is installed in another project it doesn't have issues). Does it make sense to add this while we are at it?

``` suggestion
          fi

  compatibility-test:
    name: Compatibility test
    needs: prepare
    runs-on: ubuntu-latest
    strategy:
      matrix:
        node-version: [22.x, 24.x, 26.x]
    steps:
      - name: Checkout and setup environment
        uses: MetaMask/action-checkout-and-setup@v3
        with:
          is-high-risk-environment: false
          persist-credentials: false
          node-version: ${{ matrix.node-version }}
      - name: Install dependencies via Yarn
        run: rm yarn.lock && YARN_ENABLE_IMMUTABLE_INSTALLS=false yarn
      - run: yarn test
      - name: Restore lockfile
        run: git restore yarn.lock
      - name: Require clean working directory
        shell: bash
        run: |
          if ! git diff --exit-code; then
            echo "Working tree dirty at end of job"
            exit 1
          fi
```

.github/workflows/create-release-pr.yml
mcmire, line 1: It looks like there have been other changes to `create-release-pr.yml` made since this repo was created, is it worth overwriting the file completely using https://github.com/MetaMask/metamask-module-template/blob/main/.github/workflows/create-release-pr.yml?

.github/workflows/main.yml
mcmire, line 1: It looks like `main.yml` in the module template has more changes, notably it passes `SLACK_WEBHOOK_URL` to `publish-release` so that @MetaMask/npm-publishers will be notified when there is a new release of `eth-sig-util`. Is it worth copying this file completely? https://github.com/MetaMask/metamask-module-template/blob/main/.github/workflows/main.yml

.github/workflows/publish-release.yml
```diff
@@ -1,118 +1,140 @@
 name: Publish Release
 
 on:
-  push:
-    branches: [main]
+  workflow_call:
+    secrets:
+      NPM_TOKEN:
+        required: false
```
mcmire, line 7: In module template this file accepts a `SLACK_WEBHOOK_URL` and then uses it in the `publish-npm-dry-run` job, thoughts on adding that here?

``` suggestion
        required: false
      SLACK_WEBHOOK_URL:
        required: true
```

.github/workflows/publish-release.yml
```diff
@@ -1,118 +1,140 @@
...
-          key: ${{ github.sha }}
-        # Set `ignore-scripts` to skip `prepublishOnly` because the release was built already in the previous job
-      - run: npm config set ignore-scripts true
+          name: publish-release-artifacts-${{ github.sha }}
       - name: Dry Run Publish
-        # omit npm-token token to perform dry run publish
-        uses: MetaMask/action-npm-publish@v1
+        uses: MetaMask/action-npm-publish@v6
+        with:
+          dry-run: true
```
mcmire, line 56: Thoughts on adding?

``` suggestion
          dry-run: true
          slack-webhook-url: ${{ secrets.SLACK_WEBHOOK_URL }}
          subteam: S042S7RE4AE # @metamask-npm-publishers

.github/workflows/build-lint-test.yml
OGPoyraz: Thanks @mcmire , added this

.github/workflows/main.yml
OGPoyraz: Done, copied `main.yml` from the template.

.github/workflows/publish-release.yml
OGPoyraz: Applied

.github/workflows/publish-release.yml
OGPoyraz: Applied 👍

Mrtenz (approved):

PR #426, opened 2026-09-01
ci: use OIDC token-exchange for gh-pages deploys

OGPoyraz:
## Summary

Docs deploys to `gh-pages` were failing (GH013 ruleset rejection) because #421 switched the deploy token to `GITHUB_TOKEN`, which isn't allowed to bypass the org's push-protection ruleset on `gh-pages`.

Fixes it by switching to OIDC token-exchange (same pattern as [`MetaMask/snaps`](https://github.com/MetaMask/snaps/blob/main/.github/workflows/publish-github-pages.yml)): each deploy job exchanges its OIDC identity for a short-lived `contents: write` token via `MetaMask/github-tools/.github/actions/get-token@v1`, instead of using a stored secret.

## Why this approach over reverting to the old `PUBLISH_DOCS_TOKEN` secret

That secret still exists and would also fix the immediate issue, but it's been unused/unrotated since #421 — exactly the kind of staleness that let this regression go unnoticed. 

## Validation

- `actionlint v1.7.12`: 0 errors on all changed files
- Matches `snaps`' working implementation exactly (permissions, `get-token` step, `personal_token` wiring)
- `vars.TOKEN_EXCHANGE_URL` confirmed already inherited by this repo at the org level

## Caveat

Can't verify from the API whether the token-exchange service's server-side trust policy already allows `eth-sig-util` to bypass the `gh-pages` ruleset (it evidently does for `snaps`). If not yet configured, the `Get access token` step will fail with a clear error rather than a silent bad push.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -124,3 +124,4 @@ jobs:
     secrets:
       NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
       SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+      PUBLISH_DOCS_TOKEN: ${{ secrets.PUBLISH_DOCS_TOKEN }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -124,4 +124,3 @@ jobs:
     secrets:
       NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
       SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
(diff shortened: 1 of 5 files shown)


mcmire (approved):
This looks good, as long as a policy for the new token is present in `token-exchange-service`.

mcmire:
> Can't verify from the API whether the token-exchange service's server-side trust policy already allows `eth-sig-util` to bypass the `gh-pages` ruleset (it evidently does for `snaps`).

This is something that we had to explicitly ask TechOps to do, so you may have to do the same.

OGPoyraz:
Thank you @mcmire ! 

Created a PR to add this package into `token-exchange-service` here: https://github.com/consensys-vertical-apps/token-exchange-service/pull/170

I will separately ask TechOps for the gh-pages ruleset.

Until all resolved I will keep this open

